PR #17, opened 2025-08-04
Fix #10: CA Connector

banterCZ:
Implement EJBCA connector.

PR is ready to review, but still has to be tested end-to-end against the EJBCA instance.

PR #19, opened 2025-08-07
Fix #18: Create initial DB structure

### LINKED ISSUE #18: Database schema
### Description

Server needs database for storing some data (CSR, document etc.)

### Acceptance criteria

It is possible to run migration via Liquibase for all supported databases - PostgreSQL, MS SQL, Oracle

### Technical specification

Technical description for Signer Cloud https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/Architecture.md
Best practice rules for DB https://github.com/wultra/development-internal/wiki#database-best-practices

### QA specification

_No response_

michal-rozehnal-w:
The database documentation is created using PlantUML, and a Liquibase changeset is provided for database initialization.
It also adds Spring Data JDBC DAO classes.


michal-rozehnal-w:
I used following commands for testing the changeset with docker:
```
docker run --name postgres-liquibase -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=signer_cloud_server -p 5432:5432 -d postgres:16
liquibase --url="jdbc:postgresql://localhost:5432/signer_cloud_server" --username=postgres --password=postgres --changeLogFile=docs/db/changelog/changesets/signer-cloud-server/db.changelog-module.xml update-testing-rollback
```

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,169 @@
...
+            <column name="file_name" type="varchar(255)">
+                <constraints nullable="false" />
+            </column>
+            <column name="file_size" type="integer">
+                <constraints nullable="false" />
+            </column>
+            <column name="document_content_id" type="bigint">
+                <constraints nullable="false" foreignKeyName="sc_document_document_content_id_fk" referencedTableName="sc_document_content" referencedColumnNames="id" />
+            </column>
+            <column name="hash" type="varchar(255)">
```
michal-rozehnal-w, line 134: I wasn't sure about the `varchar` type here, `clob` could also be used, depending on the value size. Same goes for the `signature` column.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,169 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.9.xsd">
```
banterCZ: Nitpick: better to use `https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd`

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,169 @@
...
+    <changeSet id="1" logicalFilePath="signer-cloud-server/1.0.x/20250807-init-db.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <sequenceExists sequenceName="sc_signer_seq"/>
+            </not>
+        </preConditions>
+        <comment>Create a new sequence sc_signer_seq</comment>
+        <createSequence sequenceName="sc_signer_seq" startValue="1" incrementBy="1" cacheSize="20"/>
+        <rollback>
+            <dropSequence sequenceName="sc_signer_seq"/>
```
banterCZ: Isn't this rollback present by default?

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,169 @@
...
+        <comment>Create a new table sc_signer</comment>
+        <createTable tableName="sc_signer">
+            <column name="id" type="bigint">
+                <constraints primaryKey="true" />
+            </column>
+            <column name="timestamp_created" type="timestamp(6)">
+                <constraints nullable="false" />
+            </column>
+            <column name="timestamp_last_updated" type="timestamp(6)">
+                <constraints nullable="false" />
```
banterCZ: 🤔 Not sure whether to have _updated_ nullable for the fresh records or not.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,169 @@
...
+            </column>
+            <column name="user_id" type="varchar(255)">
+                <constraints nullable="false" />
+            </column>
+            <column name="csr" type="clob">
+                <constraints nullable="false" />
+            </column>
+            <column name="certificate" type="clob" />
+            <column name="timestamp_certificate_expiration" type="timestamp(6)" />
+            <column name="signer_status" type="integer">
```
banterCZ: Isn't it better to have an enum string value instead?

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,169 @@
...
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="sc_signer_timestamp_certificate_expiration_idx"/>
+            </not>
+        </preConditions>
+        <comment>Create a new index sc_signer_timestamp_certificate_expiration_idx</comment>
+        <createIndex tableName="sc_signer" indexName="sc_signer_timestamp_certificate_expiration_idx">
+            <column name="timestamp_certificate_expiration" />
+        </createIndex>
+        <rollback>
```
banterCZ: ditto

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,169 @@
...
+
+    <!-- sc_document_content table -->
+    <changeSet id="4" logicalFilePath="signer-cloud-server/1.0.x/20250807-init-db.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <sequenceExists sequenceName="sc_document_content_seq"/>
+            </not>
+        </preConditions>
+        <comment>Create a new sequence sc_document_content_seq</comment>
+        <createSequence sequenceName="sc_document_content_seq" startValue="1" incrementBy="1" cacheSize="20"/>
```
banterCZ: 🤔 Not sure about cacheSize, if we will use Spring Data. I have to read some details about it.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
banterCZ: @romanstrobl Any recommendation, whether we should stick with `varchar(255)` or prefer `text`?

banterCZ (commented):
A very promissing draft, let's continue.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
michal-rozehnal-w: Updated

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
michal-rozehnal-w: Good catch! It is, I checked this [documentation](https://docs.liquibase.com/workflows/liquibase-community/automatic-custom-rollbacks.html), chapter Auto rollback Change Types and removed rollback for operations which supports automatic rollback (which means actually all of them).

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
michal-rozehnal-w: I was thinking we can set it during record creation (as we do it for `timestamp_created`). However I checked our other repositories and the `timestamp_last_update` values are nullable. In order to be consistent I made it nullable also here.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
michal-rozehnal-w: Removed

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
michal-rozehnal-w: Good question. I will also look into it.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
michal-rozehnal-w: I used number because it should be more efficient in terms of storage space. On the other hand, it is less readable, and the order of enum values can't be changed. I don't have a strong opinion about it—if you recommend using strings, I'm happy to adjust.

banterCZ: Personally, I prefer String. Historically, we have a number in PAS, and I do not like it. Yes, it saves some storage, but I vote for better data readability.

michal-rozehnal-w: OK, I changed it to `VARCHAR(50)`. We can tweak the size if needed.

docs/db/changelog/db.changelog-master.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                                       http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.10.xsd">
```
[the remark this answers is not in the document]
michal-rozehnal-w: Updated

docs-private/cloud_signer_server_database_schema.puml
```diff
@@ -0,0 +1,53 @@
+@startuml
+hide stereotypes
+hide circle
+
+class sc_signer {
+    ID bigint PK
+    timestamp_created timestamp(6) NN
+    timestamp_last_updated timestamp(6) NN
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed by removing `NN` from the documentation.

docs-private/cloud_signer_server_database_schema.puml
```diff
@@ -0,0 +1,53 @@
...
+    csr CLOB NN
+    certificate CLOB
+    timestamp_certificate_expiration timestamp(6)
+    signer_status integer NN (ACTIVE/BLOCKED/REMOVED/REVOKED)
+}
+
+class sc_document {
+    ID bigint PK
+    timestamp_created timestamp(6) NN
+    timestamp_last_updated timestamp(6) NN
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed by removing `NN` from the documentation.

docs-private/cloud_signer_server_database_schema.puml
```diff
@@ -0,0 +1,53 @@
...
+    document_id varchar(255) U NN
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed by adding `unique` constraint for the column to the changeset

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,144 @@
...
+
+    <changeSet id="2" logicalFilePath="signer-cloud-server/1.0.x/20250807-init-db.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="sc_signer"/>
+            </not>
+        </preConditions>
+        <comment>Create a new table sc_signer</comment>
+        <createTable tableName="sc_signer">
+            <column name="id" type="bigint" defaultValueComputed="nextval('sc_signer_seq')">
```
michal-rozehnal-w: Spring Data JDBC doesn't support sequence-based ID generation like JPA's `SequenceGenerator` (at least as far as I know). That's why the default value is defined at the database level.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/dao/Document.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.data.annotation.Id;
+import org.springframework.data.relational.core.mapping.Table;
+
+import java.time.Instant;
+
+/**
+ * Data Access Object for <code>sc_document</code> table.
+ */
+@Getter
+@Builder
```
michal-rozehnal-w: I can modify it now, but we’ll definitely need `toBuilder = true` later to modify fields. We can also use `SuperBuilder`.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
banterCZ: Have you read this documentation? https://docs.spring.io/spring-data/relational/reference/jdbc/sequences.html

signer-cloud-server/pom.xml
```diff
@@ -35,6 +35,17 @@
...
 
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-jdbc</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <version>42.7.3</version>
```
banterCZ: Please do not use the explicit version if defined by Spring Boot, which provides opinionated dependencies and their versions.

```suggestion
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/SignerCloudServerApplication.java
```diff
@@ -0,0 +1,12 @@
+package com.wultra.signercloud.server;
```
banterCZ: Missing copyright. Unfortunately, my PR is not yet merged to rebase.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/dao/Document.java
```diff
@@ -0,0 +1,45 @@
+package com.wultra.signercloud.server.dao;
```
banterCZ: ditto

signer-cloud-server/src/main/java/com/wultra/signercloud/server/dao/Signer.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.relational.core.mapping.Table;
+
+import java.time.Instant;
+
+/**
+ * Data Access Object for <code>sc_signer</code> table.
+ */
```
banterCZ: Missing `@author` attribute.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/repository/DocumentRepository.java
```diff
@@ -0,0 +1,10 @@
+package com.wultra.signercloud.server.repository;
+
+import com.wultra.signercloud.server.dao.Document;
+import org.springframework.data.repository.CrudRepository;
+
+/**
+ * Repository for accessing the <code>sc_document</code> table.
```
banterCZ: Nitpick: If the table name changes, you have to modify too many places. I would rather reference a Java class, something like
`Repository for {@link Document}.`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/repository/SignerRepository.java
```diff
@@ -0,0 +1,10 @@
+package com.wultra.signercloud.server.repository;
```
banterCZ: Yes, this is a working old-school layers approach. I know that we use it in our existing repos, but I would like to abandon it. We do not have to directly use Spring Modulit https://docs.spring.io/spring-modulith/reference/fundamentals.html but let's be inspired at least. I tried to demonstrate it in EJBCA package, see https://github.com/wultra/signer-cloud/pull/17/files or you may check out Malwarelytics 2.0 https://github.com/wultra/malwarelytics-server

signer-cloud-server/src/main/java/com/wultra/signercloud/server/status/DocumentStatus.java
```diff
@@ -0,0 +1,16 @@
+package com.wultra.signercloud.server.status;
+
+/**
+ * Enum representing the status of a document in the signing process.
+ *
+ * <ul>
+ *   <li>WAITING - Document is uploaded and is waiting for signature</li>
```
banterCZ: 🤔 Maybe it could be specified as a JavaDoc over the particular enum value.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -0,0 +1,9 @@
+spring.datasource.url=jdbc:postgresql://localhost:5432/testdb
+spring.datasource.username=postgres
+spring.datasource.password=postgres
+spring.datasource.driver-class-name=org.postgresql.Driver
+
+spring.sql.init.mode=never
+spring.data.jdbc.repositories.enabled=true
+
+logging.level.org.springframework.jdbc.core=DEBUG
```
banterCZ: Let's move such fine-grained logging configuration to `application-dev.properties`. You may turn on this profile for local development.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -0,0 +1,9 @@
+spring.datasource.url=jdbc:postgresql://localhost:5432/testdb
```
banterCZ: I believe that we should have a different default value than `testdb`.

signer-cloud-server/src/test/java/com/wultra/security/signer/server/DaoTest.java
```diff
@@ -0,0 +1,86 @@
...
+ * Class for Spring JDBC DAO tests.
+ *
+ * It is disabled because it requires a running database.
+ * You can set up a database locally by these commands (from the project root directory):
+ * <pre> <code>
+ * docker run --name postgres-liquibase -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=testdb -p 5432:5432 -d postgres:16
+ * liquibase --url="jdbc:postgresql://localhost:5432/testdb" --username=postgres --password=postgres --changeLogFile=docs/db/changelog/changesets/signer-cloud-server/db.changelog-module.xml update
+ * </code> </pre>
+ */
+@Disabled
```
banterCZ: May we have tests over **h2** engine? Moreover, the configuration should be defined in `application-test.properties`

signer-cloud-server/src/test/java/com/wultra/security/signer/server/DaoTest.java
```diff
@@ -0,0 +1,86 @@
...
+    private SignerRepository signerRepository;
+
+    @Autowired
+    private DocumentRepository documentRepository;
+
+    @Autowired
+    private DocumentContentRepository documentContentRepository;
+
+    /**
+     * Tests all DAO save operations.
```
banterCZ: A good start, but I would rather see more granular test classes. I mean having a test class per a repository.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
michal-rozehnal-w: No, I completely missed that. It’s exactly what we need, and locally it works correctly. Thank you!

signer-cloud-server/pom.xml
michal-rozehnal-w: Removed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/SignerCloudServerApplication.java
michal-rozehnal-w: ype, we will have a merge conflict here (plus in the properties file). I added the copyright and also updated my default templates in IntelliJ, so it’s there by default for any new class.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/dao/Document.java
michal-rozehnal-w: Updated

signer-cloud-server/src/main/java/com/wultra/signercloud/server/dao/Signer.java
michal-rozehnal-w: Added and I also set template for author in IntelliJ

signer-cloud-server/src/main/resources/application-dev.properties
```diff
@@ -0,0 +1,6 @@
+spring.datasource.url=jdbc:postgresql://localhost:5432/signer_cloud_server
+spring.datasource.username=postgres
+spring.datasource.password=postgres
+spring.datasource.driver-class-name=org.postgresql.Driver
```
banterCZ: 🤔 If using JPA, Spring is able to guess the driver from the connection string. Doesn't it work here, too?

signer-cloud-server/src/test/java/com/wultra/security/signer/server/DaoTest.java
```diff
@@ -0,0 +1,88 @@
...
+ * It is disabled because it requires a running database.
+ * You can set up a database locally by these commands (from the project root directory):
+ * <pre> <code>
+ * docker run --name postgres-liquibase -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=testdb -p 5432:5432 -d postgres:16
+ * liquibase --url="jdbc:postgresql://localhost:5432/testdb" --username=postgres --password=postgres --changeLogFile=docs/db/changelog/changesets/signer-cloud-server/db.changelog-module.xml update
+ * </code> </pre>
+ */
+@Disabled
+@SpringBootTest(classes = SignerCloudServerApplication.class)
+@ActiveProfiles("dev")
```
banterCZ: I would expect the profile `test`. Maybe I confused you, because I am using `dev` profile for the integration test.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/repository/DocumentRepository.java
michal-rozehnal-w: Good point! Although IntelliJ nowadays has very good support for renaming/refactoring, using the link is better. Additionally, it decouples classes from the DB representation, since we reference the DTO and not the table.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/repository/SignerRepository.java
michal-rozehnal-w: Thank you for broadening my knowledge. I split it by `document` and `signer`, which makes the most sense, but I’m open to any other ideas.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/status/DocumentStatus.java
michal-rozehnal-w: You are right, it’s better since you can just see the enum value and don’t have to read the documentation for the entire class. Updated.

signer-cloud-server/src/main/resources/application.properties
michal-rozehnal-w: Moved to `application-dev.properties`

signer-cloud-server/src/main/resources/application.properties
michal-rozehnal-w: You are absolutely right, this was just the first kick-off. I renamed it to `signer_cloud_server`

signer-cloud-server/src/test/java/com/wultra/security/signer/server/DaoTest.java
michal-rozehnal-w: Sorry, this is not a final version, I just used `dev` in order to test it locally. I will do it properly now (break the test into smaller ones per repository, add test properties file etc.)

michal-rozehnal-w: All test specific values are in `application-test.properties` now

signer-cloud-server/src/main/resources/application-dev.properties
michal-rozehnal-w: You are right, this is common feature in Spring, I removed the property from all `*.properties` files

signer-cloud-server/src/test/java/com/wultra/security/signer/server/DaoTest.java
michal-rozehnal-w: Done. We have H2 where changes from Liquibase are applied and then tests are executed.

signer-cloud-server/src/test/java/com/wultra/security/signer/server/DaoTest.java
michal-rozehnal-w: Done, each repository has its own test

signer-cloud-server/src/test/java/com/wultra/security/signer/server/DaoTest.java
```diff
@@ -0,0 +1,88 @@
...
+     * Tests all DAO save operations.
+     */
+    @Test
+    public void testSave() {
+        // create a new signer
+        var signer = Signer.builder()
+                .timestampCreated(Instant.now())
+                .signerExternalId("signer123")
+                .userId("user123")
+                .csr("crs123")
```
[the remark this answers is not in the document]
michal-rozehnal-w: I put value into constant and it is correctly spelled.

signer-cloud-server/src/main/resources/application-dev.properties
```diff
@@ -0,0 +1,6 @@
+spring.datasource.url=jdbc:postgresql://localhost:5432/signer_cloud_server
+spring.datasource.username=postgres
+spring.datasource.password=postgres
```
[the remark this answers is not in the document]
michal-rozehnal-w: The DB name for tests was updated to `signer_cloud_server`

signer-cloud-server/src/test/resources/application-test.properties
```diff
@@ -1 +1,10 @@
 signer-cloud.server.ejbca.rest-client-configuration.certificate-auth-enabled=false
+
+spring.datasource.url=jdbc:h2:mem:signer_cloud_server;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE;CASE_INSENSITIVE_IDENTIFIERS=TRUE
+spring.datasource.username=sa
+spring.datasource.password=
+
+spring.liquibase.change-log=file:../docs/db/changelog/db.changelog-master.xml
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is intentional because the Liquibase files are located outside of `src`, so the `classpath` won’t work. I don’t think we will change the project structure significantly, and if we do, this value is applied only for tests. In the worst case, the tests will fail, which would be an easy fix. Based on that, I don’t see it as an issue.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,144 @@
...
+            <column name="signer_external_id" type="varchar(255)">
+                <constraints nullable="false" unique="true"  uniqueConstraintName="sc_signer_external_id_uk" />
+            </column>
+            <column name="user_id" type="varchar(255)">
+                <constraints nullable="false" />
+            </column>
+            <column name="csr" type="clob">
+                <constraints nullable="false" />
+            </column>
+            <column name="certificate" type="clob" />
```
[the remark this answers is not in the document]
michal-rozehnal-w: Currently, we use Postgres and H2, and for both, `clob` is a valid type. In the future, we may add Oracle and MS SQL, for which this value is still acceptable. I don’t think it’s necessary to introduce a property for it.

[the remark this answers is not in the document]
michal-rozehnal-w: Same as previous

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
michal-rozehnal-w: As we discussed, I set the default value to `50` and did some testing locally with Postgres in Docker. For each iteration, I started the app, inserted a Signer, DocumentContent, and Document, and printed their IDs. Then I tore down the app.
```
Iteration 1: Signer ID=1, DocumentContent ID=1, Document ID=1
delay few seconds
Iteration 2: Signer ID=51, DocumentContent ID=51, Document ID=51
delay 3 minutes
Iteration 3: Signer ID=101, DocumentContent ID=101, Document ID=101
```

michal-rozehnal-w: Additional note: If more records are stored during a single app run, then the IDs are in sequence (1, 2, 3 …), so there is no gap of `50` (the size of cache).

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,144 @@
...
+
+    <!-- sc_signer table -->
+    <changeSet id="1" logicalFilePath="signer-cloud-server/1.0.x/20250807-init-db.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <sequenceExists sequenceName="sc_signer_seq"/>
+            </not>
+        </preConditions>
+        <comment>Create a new sequence sc_signer_seq</comment>
+        <createSequence sequenceName="sc_signer_seq" startValue="1" incrementBy="1" cacheSize="50"/>
```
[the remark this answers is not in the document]
michal-rozehnal-w: The `50` should be fine (it is default value used by Spring Boot). We can adjust the value any time in future by another changeset

signer-cloud-server/src/test/resources/application-test.properties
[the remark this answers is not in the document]
michal-rozehnal-w: This is on purpose, I explained it earlier (the `/docs` is outside of `/src`)

docs-private/cloud_signer_server_database_schema.puml
```diff
@@ -0,0 +1,53 @@
...
+class sc_signer {
+    ID bigint PK
+    timestamp_created timestamp(6) NN
+    timestamp_last_updated timestamp(6)
+    signer_external_id varchar(255) U NN
+    user_id varchar(255) NN
+    csr CLOB NN
+    certificate CLOB
+    timestamp_certificate_expiration timestamp(6)
+    signer_status integer NN (ACTIVE/BLOCKED/REMOVED/REVOKED)
```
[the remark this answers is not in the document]
michal-rozehnal-w: Good catch, fixed.

docs-private/cloud_signer_server_database_schema.puml
```diff
@@ -0,0 +1,53 @@
...
+    timestamp_last_updated timestamp(6)
+    document_id varchar(255) U NN
+    signer_id bigint FK
+    external_id varchar(255)
+    document_name varchar(255) NN
+    file_name varchar(255) NN
+    file_size integer NN
+    document_content_id bigint FK
+    hash varchar(255) NN
+    document_status integer NN (WAITING/REJECTED/SIGNED)
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

docs-private/cloud_signer_server_database_schema.puml
banterCZ, line 1: Should be named as `signer_cloud` instead of `cloud_signer`?

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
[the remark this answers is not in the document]
banterCZ: Exactly. You may see the mapping at https://docs.liquibase.com/concepts/data-type-handling.html

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
```diff
@@ -0,0 +1,144 @@
...
+                <constraints nullable="false" foreignKeyName="sc_document_signer_id_fk" referencedTableName="sc_signer" referencedColumnNames="id" />
+            </column>
+            <column name="external_id" type="varchar(255)" />
+            <column name="document_name" type="varchar(255)">
+                <constraints nullable="false" />
+            </column>
+            <column name="file_name" type="varchar(255)">
+                <constraints nullable="false" />
+            </column>
+            <column name="file_size" type="integer">
```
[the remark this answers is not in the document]
banterCZ: I guess that we do not want to sign 2GB PDF files anyway.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/Document.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Data Access Object for the <code>sc_document</code> table.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@Getter
+@Builder
+@Table("sc_document")
+public class Document {
```
banterCZ, line 36: 🤔 I guess that we may introduce `equals`, `hashcode`, and `toString` later.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/Signer.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Id
+    @Sequence("sc_signer_seq")
+    private long id;
+
+    private Instant timestampCreated;
+
+    private Instant timestampLastUpdated;
+
+    private String signerExternalId;
```
banterCZ: 🤔 Not `externalSignerId`?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/Signer.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private String userId;
+
+    private String csr;
+
+    private String certificate;
+
+    private Instant timestampCertificateExpiration;
+
+    private SignerStatus signerStatus;
```
banterCZ: 🤔 Why not only `status`? I mean `SignerStatus status;`.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/Document.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Id
+    @Sequence("sc_document_seq")
+    private long id;
+
+    private Instant timestampCreated;
+
+    private Instant timestampLastUpdated;
+
+    private String documentId;
```
banterCZ, line 46: Spring data is missing a link from Java code. What about keeping this information in JavaDoc, something like.
```suggestion
    /**
    * {@link Document#id}
    */
    private String documentId;
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentContent.java
```diff
@@ -0,0 +1,41 @@
...
+package com.wultra.signercloud.server.document;
+
+import lombok.Builder;
+import lombok.Getter;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.relational.core.mapping.Sequence;
+import org.springframework.data.relational.core.mapping.Table;
+
+/**
+ * Data Access Object for the <code>sc_document_content</code> table.
```
banterCZ: You may also use `{@code sc_document_content}` if you want to.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentStatus.java
```diff
@@ -0,0 +1,40 @@
...
+package com.wultra.signercloud.server.document;
+
+/**
+ * Enum representing the status of a {@link Document} in the signing process.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public enum DocumentStatus {
+    /**
+     * Document is uploaded and is waiting for signature
```
banterCZ: Probably add a point at the end of the sentence.
```suggestion
     * Document is uploaded and is waiting for signature.
```

signer-cloud-server/src/main/resources/application-dev.properties
```diff
@@ -8,3 +8,9 @@ signer-cloud.server.ejbca.rest-client-configuration.key-store-password=
 signer-cloud.server.ejbca.rest-client-configuration.key-alias=
 signer-cloud.server.ejbca.rest-client-configuration.key-password=
 signer-cloud.server.ejbca.rest-client-configuration.key-store-bytes=
+
+spring.datasource.url=jdbc:postgresql://localhost:5432/signer_cloud_server
+spring.datasource.username=postgres
+spring.datasource.password=postgres
```
banterCZ: I suggest using the dev database without the password.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -8,3 +8,10 @@ signer-cloud.server.ejbca.rest-client-configuration.key-store-password=${EJBCA_K
 signer-cloud.server.ejbca.rest-client-configuration.key-alias=${EJBCA_KEY_ALIAS:}
 signer-cloud.server.ejbca.rest-client-configuration.key-password=${EJBCA_KEY_PASSWORD:}
 signer-cloud.server.ejbca.rest-client-configuration.key-store-bytes=${EJBCA_KEYSTORE_BYTES:}
+
+spring.datasource.url=${DB_URL:}
```
banterCZ: What about
```suggestion
spring.datasource.url=${DB_URL:jdbc:postgresql://localhost:5432/signer_cloud_server}
```

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -8,3 +8,10 @@ signer-cloud.server.ejbca.rest-client-configuration.key-store-password=${EJBCA_K
...
+spring.datasource.username=${DB_USERNAME:}
+spring.datasource.password=${DB_PASSWORD:}
+
+spring.sql.init.mode=never
+spring.data.jdbc.repositories.enabled=true
```
banterCZ: Not a strong opinion, but do you know the equivalent annotation `@EnableJdbcRepositories`? Have you considered that?

banterCZ (approved):
Good job! I have added several less severe comments to consider.

docs-private/cloud_signer_server_database_schema.puml
michal-rozehnal-w: Well spotted! Updated.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/Document.java
michal-rozehnal-w: Yes, we can add them at any time later. I create them when I really need them, but it’s purely a personal preference.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/Signer.java
michal-rozehnal-w: I was thinking about the name and having the `external` first is probably better, because it immediately indicates it is not our internal id. I updated the name

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/Signer.java
michal-rozehnal-w: Right, since this is in `Signer` class the `signer` prefix is redundant. I did same update for `Document` status.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/Document.java
michal-rozehnal-w: Good point, I added links to all binding ids

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentContent.java
michal-rozehnal-w: Thank you for this suggestion. I used it.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentStatus.java
michal-rozehnal-w: Yes, updated.

signer-cloud-server/src/main/resources/application-dev.properties
michal-rozehnal-w: Removed

signer-cloud-server/src/main/resources/application.properties
michal-rozehnal-w: Sounds good to me. Updated

signer-cloud-server/src/main/resources/application.properties
michal-rozehnal-w: Thank you, I wasn't aware of this annotation. I will consider using it next time.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250807-init-db.xml
romanstrobl: `varchar(255)` is safe even for 1024-bit hashes and we are going to use much smaller bit sizes. For performance reasons it is better to stick to `varchar` over `text`.

PR #33, opened 2025-08-13
Fix #32: Spring Boot Custom Banner

### LINKED ISSUE #32: Spring Boot Custom Banner
### Task specification

_No response_

banterCZ:


signer-cloud-server/src/main/resources/banner.txt
```diff
@@ -0,0 +1,10 @@
+  ____  _                          ____ _                 _
```
michal-rozehnal-w, line 1: We have `spring.application.name=signer-cloud-server`, but banner for `Signer Cloud`. Is it on purpose truncated or do we want to add the `Server` word?

banterCZ: A good point. I tried it, but `Signer Cloud Server` seems too long to me. When we deploy more signer artifacts, we may change the banner.

michal-rozehnal-w (approved):

PR #36, opened 2025-08-13
Fix #24: Create Docker

### LINKED ISSUE #24: Create Docker
### Task specification

Create Docker image and configure GHA

banterCZ:
Heavily inspired by _user-data-store_ build.


korbelm (changes requested):
please use pipeline and setup from mwl 2.0

michal-rozehnal-w (approved):

deploy/docker-entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+liquibase --headless=true --log-level=INFO --changeLogFile="$LB_HOME/db/changelog/db.changelog-master.xml" \
+  --username="${DB_USERNAME:-}" \
+  --password="${DB_PASSWORD:-}" \
+  --url="${DB_URL}" \
+  update
+
+java "${JAVA_OPTS:-}" -jar signer-cloud-server.war
```
jnpsk: 🤔  IMHO it should not be quoted here, as it means that all the java opts will be passed as single argument.

banterCZ: Shame on Copilot, he suggested this change but when challenged:
> If JAVA_OPTS contains multiple options, wrapping in quotes will pass them as one long argument, rather than as separate options. For example, -Xmx2g -Dfoo=bar would be passed as one argument, which Java does not expect.

jnpsk (approved):

korbelm (approved):
publish pipeline looks OK

deploy/docker-entrypoint.sh
korbelm: its common practice for bash script to "not split string if there is space in value" but thats what we dont want here

Issue #27, opened 2025-08-12
Make possible to configure security using OAuth2


banterCZ:
### Task specification

Mind to specify that in OpenAPI as well.

PR #38, opened 2025-08-14
Fix #2: REST API Create New Signer

### LINKED ISSUE #2: REST - Create New Signer
### Description

Implement POST method for creating a signer (entity that enrolls for certificate).

It's recommended to implement signer management in following order:
- create
- change status / details

### Acceptance criteria

### Technical specification

[Create New Signer](https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/CloudSigner-API.md#create-new-signer)

- we need CSR at the input (PEM encoded PKCS10 CSR as one line string with line endings `\n`)
- CSR should be synchronously passed to the CA (EJBCA) to enroll for certificate... API method TBC
- CSR needs to be stored for certificate renewal
- certificate needs to be stored for document signing
- ~~CSR and certificate need to be stored securely~~ After the discussion we had, we don’t need to store it securely since both values are public information.

### QA specification
- test method to create new signer
- verify that certificate was created on CA, test also wrong state (CA is not available or is wrongly configured)
- test dependent behavior - certificate renewal job

michal-rozehnal-w:
Endpoint for create new signer.
Successfully tested locally with our internal EJBCA instance.


signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -0,0 +1,65 @@
...
+                    ),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid input data"
+                    )
+            }
+    )
+    @PostMapping
+    public ResponseEntity<SignerResponse> create(@Valid @RequestBody CreateSignerRequest createSignerRequest) {
+        final var response = signerService.createSigner(createSignerRequest);
```
michal-rozehnal-w: Personally, I prefer thin controllers that simply call a service. The business logic is in the service, and the controller serves as configuration for API paths, roles, OpenAPI documentation, etc.
Please let me know if we want to use a different approach.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerResponse.java
```diff
@@ -0,0 +1,29 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.wultra.signercloud.server.signer;
+
+/**
+ * Common response for operations related to {@link Signer}.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public record SignerResponse(
```
michal-rozehnal-w: On purpose the name is `SignerResponse` and not `CreateSignerResponse`, because Change Signer Status operation has exactly same response structure and I was thinking we can use this response there.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,82 @@
...
+    private void createSignerWithCertificate(CreateSignerRequest request) throws RestClientException, CertificateException, IOException {
+        final var externalSignerId = request.signerId();
+
+        var isRegistrationActive = powerAuthService.isRegistrationActive(externalSignerId);
+        if (!isRegistrationActive) {
+            throw new IllegalStateException("Signer registration is not active for external signer ID: " + externalSignerId);
+        }
+
+        final var csr = request.csr();
+        var x509Certificate = ejbcaService.enrollCertificate(csr);
```
michal-rozehnal-w: In case of an error, I presume the activation check is an idempotent action, but I am not sure about the certificate creation. If the call is successful and we obtain a certificate, but an exception is thrown afterward, we do not save it into the database. My concern is about having some unlinked certificates in EJBCA. I will review the documentation and try this scenario locally.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,82 @@
...
+                .timestampCreated(Instant.now())
+                .externalSignerId(request.signerId())
+                .userId(request.userId())
+                .csr(csr)
+                .certificate(certificate)
+                .timestampCertificateExpiration(certificateExpiration)
+                .status(SignerStatus.ACTIVE)
+                .build();
+
+        signerRepository.save(signer);
```
michal-rozehnal-w, line 97: This is somewhat related to the previous comment. A user can renew a certificate manually, which means calling this endpoint with the same `userId` and `signerId`, but a different `csr` (although it can actually be the same in case of a retry for example). In this case, I think we should just update the existing record (set `timestamp_last_update`) rather than create a new one.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
banterCZ: Great, I totally agree.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Controller for {@link Signer} operations.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@RestController
+@RequestMapping("api/signers")
```
banterCZ, line 38: ℹ️ We have not yet configured the security; it is ok for this PR. We will do it in a separate PR, most probably #27

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -0,0 +1,65 @@
...
+                            content = @Content(schema = @Schema(implementation = SignerResponse.class))
+                    ),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid input data"
+                    )
+            }
+    )
+    @PostMapping
+    public ResponseEntity<SignerResponse> create(@Valid @RequestBody CreateSignerRequest createSignerRequest) {
```
banterCZ: Nitpick: I am just obsessed by `final`
```suggestion
    public ResponseEntity<SignerResponse> create(@Valid @RequestBody final CreateSignerRequest createSignerRequest) {
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
banterCZ: Please add logging according to the guidelines https://github.com/wultra/development-internal/wiki/Logging-Guideline
Just a hint, templates such as `action: {}` are IMHO overhead and you may inline that.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
banterCZ: `org.springframework.http.ResponseEntity` itself is superfluous; sometimes we use `com.wultra.core.rest.model.base.response.ObjectResponse`, which is not specified in the documentation https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/CloudSigner-API.md but we may double-check with @tech-kratas

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerResponse.java
banterCZ: Ok so far, just a note, that duplicity (in general) is better than wrong abstraction, see https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Service for {@link Signer} operations.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@Service
+@AllArgsConstructor
+@Slf4j
+public class SignerService {
```
banterCZ: We should start transactions at the service layer.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private final PowerAuthService powerAuthService;
+    private final EjbcaService ejbcaService;
+    private final SignerRepository signerRepository;
+
+    public SignerResponse createSigner(CreateSignerRequest request) {
```
banterCZ: Missing JavaDoc.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,82 @@
...
+        try {
+            createSignerWithCertificate(request);
+            return new SignerResponse(SignerResponseResult.OK, null);
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
banterCZ: I agree with Copilot, `Exception` is too broad. We may introduce the exception handler as a controller advice for specific handling, if not satisfied with the default one.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,82 @@
...
+            logger.error("Exception during signer creation", e);
```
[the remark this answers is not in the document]
banterCZ: Copilot is absolutely lame and always missing `lombok.config`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,82 @@
...
+            return new SignerResponse(SignerResponseResult.FAIL, e.getMessage());
+        }
+    }
+
+    private void createSignerWithCertificate(CreateSignerRequest request) throws RestClientException, CertificateException, IOException {
+        final var externalSignerId = request.signerId();
+
+        var isRegistrationActive = powerAuthService.isRegistrationActive(externalSignerId);
+        if (!isRegistrationActive) {
+            throw new IllegalStateException("Signer registration is not active for external signer ID: " + externalSignerId);
```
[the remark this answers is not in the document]
banterCZ: On the contrary, I like your approach to add context to the exception message. However, I may stick with `externalSignerId` to improve full-text search results. Last, but not least, we should differentiate negative scenarios with some specific exceptions, not a general `IllegalStateException`.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
banterCZ: cc @tech-kratas

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
banterCZ: It is a more philosophical topic, but we may make all classes package-private. We do not need to expose `SignerService` yet, and the controller methods are exposed even if not public.

banterCZ (commented):
A very promising draft, looking forward to tests.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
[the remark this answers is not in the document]
michal-rozehnal-w: Ok, I used the suggestion from Copilot and created custom exception. I am still thinking about it, so this is probably not the final solution

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
michal-rozehnal-w: Thank you for mentioning that. As we discussed, we probably don't need it know (in this PR) and we can enable any authentication when we need to deploy this app.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
michal-rozehnal-w: Added

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
michal-rozehnal-w: Right, I haven't looked into the code from logging perspective. Thank you for the guidelines, I add the logging in one of next commits

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: I added the `@Transactional` annotation to the method. It seems better than using it at the class level because it allows precise control over that specific method. However, I don’t have a strong preference—both options are fine. Do you have a preference?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Added

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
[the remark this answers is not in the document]
michal-rozehnal-w: I added custom exception for inactive signer in powerauth. I am still thinking about the exception handling concept.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -0,0 +1,67 @@
...
+                            content = @Content(schema = @Schema(implementation = SignerResponse.class))
+                    ),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid input data"
+                    )
+            }
+    )
+    @PostMapping
+    ResponseEntity<SignerResponse> crateUpdate(@Valid @RequestBody final CreateUpdateSignerRequest requestBody) {
```
banterCZ: As Filip confirmed, we do not want to use `ObjectResponse`, but we do not need `ResponseEntity` either.
```suggestion
   SignerResponse crateUpdate(@Valid @RequestBody final CreateUpdateSignerRequest requestBody) {
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerResponse.java
michal-rozehnal-w: I agree, it can lead to the same hell as wrong inheritance. Thank you for highlighting that.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: We fixed this by sending the `userId` and `externalSignerId` in the EJBCA request. This links all issued certificates to a specific user, so in case of any action, we can work with them collectively.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: In case that `externalSignerId` already exists, we just do an update.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
michal-rozehnal-w: Added.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Yes, I like this approach and I updated the access modifier.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private Signer.SignerBuilder createSigner(String externalSignerId) {
+        return Signer.builder()
+                .timestampCreated(Instant.now())
+                .externalSignerId(externalSignerId);
+    }
+
+    private Signer.SignerBuilder updateSigner(Signer signer) {
+        return signer.toBuilder()
+                .timestampCreated(Instant.now());
```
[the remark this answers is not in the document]
michal-rozehnal-w: Good catch! Yes, we want to update `timestampLastUpdated`. I also updated the tests because it means they weren’t good enough.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

michal-rozehnal-w: Right, since we don’t need any extra manipulation of the response, I removed the wrapper.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,114 @@
...
+     * This method checks whether the registration in PowerAuth is active, then generates a certificate via the EJBCA service
+     * (based on {@link CreateUpdateSignerRequest#csr}), and finally stores the signer in the database.
+     *
+     * @param request the request containing details of signer
+     * @return {@link SignerResponse} indicating the result of the operation.
+     */
+    @Transactional
+    SignerResponse createUpdateSigner(CreateUpdateSignerRequest request) {
+        try {
+            logger.info("action: createUpdateSigner, state: initiated, userId: {}, externalSignerId: {}", request.userId(), request.signerId());
```
banterCZ: Please move this logger to the controller level. In this case, the effect is the same, but in general, it may break without logging anything at all.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerControllerIntTest.java
```diff
@@ -0,0 +1,266 @@
...
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@SpringBootTest(properties = {
+        "ejbca.rest-client.key-store-password=testPassword",
+        "ejbca.rest-client.key-alias=testAlias",
+        "ejbca.rest-client.key-password=testKeyPassword"
+})
+@AutoConfigureMockMvc
+@ActiveProfiles("test")
+class SignerControllerIntTest {
```
banterCZ, line 56: 🤔 We have not yet distinguished unit and integration tests by `Int`. We stack them on the same pile, so far.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -35,4 +35,6 @@ signer-cloud.server.document.signed.retention-period=
 management.endpoints.web.exposure.include=health,info,metrics
 management.info.git.mode=simple
 
-management.tracing.sampling.probability=1.0
\ No newline at end of file
+management.tracing.sampling.probability=1.0
+
+server.port=8090
```
banterCZ, line 40: This should be reflected in `Developer-How-To-Start.md`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -0,0 +1,68 @@
...
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid input data"
+                    )
+            }
+    )
+    @PostMapping
+    SignerResponse createUpdate(@Valid @RequestBody final CreateUpdateSignerRequest requestBody) {
+        logger.info("action: createUpdateSigner, state: initiated, userId: {}, externalSignerId: {}", requestBody.userId(), requestBody.signerId());
+        return signerService.createUpdateSigner(requestBody);
```
banterCZ: We should log success as well.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Creates a new {@link Signer} or updates an existing one if it already exists (based on {@link Signer#externalSignerId}).
+     * This method checks whether the registration in PowerAuth is active, then generates a certificate via the EJBCA service
+     * (based on {@link CreateUpdateSignerRequest#csr}), and finally stores the signer in the database.
+     *
+     * @param request the request containing details of signer
+     * @return {@link SignerResponse} indicating the result of the operation.
+     */
+    @Transactional
+    SignerResponse createUpdateSigner(CreateUpdateSignerRequest request) {
```
banterCZ:
```suggestion
    SignerResponse createUpdateSigner(final CreateUpdateSignerRequest request) {
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Moved. I did it just for the `state: initiated` message. Theoretically I can do it for results too, but it will need some if statement. It seems to me more clear to have it in service, but please let me know if the controller would be better place.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerControllerIntTest.java
michal-rozehnal-w: Yep, this is purely about team agreement. On my previous project, we used Int because it’s easy to distinguish unit and integration tests at first sight, and it’s good preparation in case we want integration tests to run in separate jobs.

signer-cloud-server/src/main/resources/application.properties
michal-rozehnal-w: You are right, I missed this step. I will also add a description of the dependencies, since it now requires PowerAuth and an EJBCA instance to operate successfully.

michal-rozehnal-w: Updated

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
michal-rozehnal-w: Moved from the service layer

docs-private/Developer-How-To-Start.md
```diff
@@ -1,17 +1,28 @@
...
 
-
 ## Signer Cloud Server
 
-
 ### Standalone Run
 
 - Use IntelliJ Idea run configuration at `../.run/SignerCloudServerApplication.run.xml`
-- Open [http://localhost:8080/actuator/health](http://localhost:8080/actuator/health) and you should get `{"status":"UP"}`
+- Open [http://localhost:8090/actuator/health](http://localhost:8080/actuator/health) and you should get `{"status":"UP"}`
```
banterCZ:
```suggestion
- Open [http://localhost:8090/actuator/health](http://localhost:8090/actuator/health) and you should get `{"status":"UP"}`
```

docs-private/Developer-How-To-Start.md
```diff
@@ -1,17 +1,28 @@
...
+
+> **Note:** The application runs on port `8090` by default to avoid conflicts with other apps in the infrastructure (for example, PowerAuth).
```
banterCZ: 🤔 I do not think that `>` as a quote is appropriate here. See Docucheck makros and consider an info or warn box https://github.com/wultra/docucheck/blob/develop/docs/Basic-Principles.md#info-boxes

docs-private/Developer-How-To-Start.md
```diff
@@ -1,17 +1,28 @@
...
 
 
 ### Database
 
+We support:
+- PostgreSQL
+- H2
```
banterCZ:
```suggestion
- H2 for test purposes
```

docs-private/Developer-How-To-Start.md
```diff
@@ -1,17 +1,28 @@
...
+
+For local development you can use docker image
+```shell
+docker run --name postgres-liquibase -e POSTGRES_USER=signer_cloud_server -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_DB=signer_cloud_server -p 5432:5432 -d postgres:16
+```
+Don't forget to run the Liquibase script (see the section below); it is not applied automatically.
+
+#### Liquibase
```
banterCZ, line 31:
```suggestion
#### Liquibase

```

docs-private/Developer-How-To-Start.md
```diff
@@ -32,28 +43,40 @@ liquibase --changelog-file=./docs/db/changelog/changesets/signer-cloud-server/db
 
 To generate SQL script run this command.
 
-
-#### PostgreSQL
```
banterCZ: Why to drop this header? What about changing it into `#####`?

docs-private/Developer-How-To-Start.md
```diff
@@ -32,28 +43,40 @@ liquibase --changelog-file=./docs/db/changelog/changesets/signer-cloud-server/db
...
-
 ```shell
+# PostgreSQL
 liquibase --changeLogFile=./docs/db/changelog/changesets/signer-cloud-server/db.changelog-module.xml --output-file=./docs/sql/postgresql/generated-postgresql-script.sql updateSQL --url=offline:postgresql
 ```
 
+## PowerAuth
+
+The PowerAuth instance checks whether the signer is active (can be used for certificate creation). \
```
banterCZ: I believe that `\` is superfluous; the markdown is able to work with line breaks. 
```suggestion
The PowerAuth instance checks whether the signer is active (can be used for certificate creation).
```

docs-private/Developer-How-To-Start.md
```diff
@@ -32,28 +43,40 @@ liquibase --changelog-file=./docs/db/changelog/changesets/signer-cloud-server/db
...
-        'https://smoke-ejbca-dev.wultra.app/ejbca/ejbca-rest-api/v1/certificate/pkcs10enroll' \
-        --cert client.crt \
-        --key client.key \
-        --json '{
-        "end_entity_profile_name": "ExampleEEP",
-        "certificate_authority_name": "ExampleCA",
-        "certificate_request": "MIICh...V8shQ== OR -----BEGIN CERTIFICATE REQUEST-----\nMIICh...V8shQ==\n-----END CERTIFICATE REQUEST-----",
-        "certificate_profile_name": "ENDUSER"
-        }'
+--data '{
```
banterCZ: I see the only `--data` but not the `curl` itself.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Updated. I also updated my IntelliJ setup so that it warns me whenever it’s possible to use `final`.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,109 @@
...
+     * This method checks whether the registration in PowerAuth is active, then generates a certificate via the EJBCA service
+     * (based on {@link CreateUpdateSignerRequest#csr}), and finally stores the signer in the database.
+     *
+     * @param request the request containing details of signer
+     * @return {@link SignerResponse} indicating the result of the operation.
+     */
+    @Transactional
+    SignerResponse createUpdateSigner(final CreateUpdateSignerRequest request) {
+        try {
+            createUpdateSignerWithCertificate(request);
```
banterCZ, line 57: ℹ️ Just highlighting, not needed to solve that in this PR. We have a similar architecture glitch elsewhere. We are calling REST in the transaction. A possible rollback does not undo the REST call and, moreover, prolongs holding database resources.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -0,0 +1,73 @@
...
+                            responseCode = "400",
+                            description = "Invalid input data"
+                    )
+            }
+    )
+    @PostMapping
+    SignerResponse createUpdate(@Valid @RequestBody final CreateUpdateSignerRequest requestBody) {
+        logger.info("action: createUpdateSigner, state: initiated, userId: {}, externalSignerId: {}", requestBody.userId(), requestBody.signerId());
+        final var result = signerService.createUpdateSigner(requestBody);
+        logger.info("action: createUpdateSigner, state: {}, errorMessage: {}",
```
banterCZ: Just a matter of taste, but probably I like `if` and two log statements more than a single log with a ternary operator. Now we would log `errorMessage: null` for success, wouldn't we?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -0,0 +1,109 @@
...
+            return new SignerResponse(SignerResponseResult.OK, null);
+        } catch (InactiveSignerException | RestClientException | CertificateException | IOException e) {
+            return new SignerResponse(SignerResponseResult.FAIL, e.getMessage());
```
banterCZ: Great, I love this functional approach, but let's make it perfect. The error object should be a specialization, like `reactor.core.publisher.Mono#error()`. Venkat describes something similar in his book [Functional Programming in Java](https://pragprog.com/titles/vsjava2e/functional-programming-in-java-second-edition), source code available at https://media.pragprog.com/titles/vsjava2e/code/vsjava2e-code.zip

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/ErrorResponse.java
```diff
@@ -0,0 +1,27 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.wultra.signercloud.server.restapi;
+
+/**
+ * Response for HTTP error.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public record ErrorResponse(String status, ErrorDetails responseObject) {}
```
banterCZ, line 25: I know that this was created based on the specification. We may reopen with @tech-kratas later on. I am just wondering why not use RFC 7807, which is supported since Spring 6, see for example
- https://dev.to/noelopez/spring-rest-exception-handling-problem-details-2hkj
- https://docs.spring.io/spring-framework/docs/6.0.0/reference/html/web.html#mvc-ann-rest-exceptions

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/ValidationExceptionHandler.java
```diff
@@ -0,0 +1,48 @@
...
+ * Handler for validation exception producing HTTP 400 response.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@ControllerAdvice
+public class ValidationExceptionHandler {
+    private static final String ERROR_STATUS = "ERROR";
+
+    @ExceptionHandler(MethodArgumentNotValidException.class)
+    public ResponseEntity<ErrorResponse> handleValidationException(MethodArgumentNotValidException ex) {
```
banterCZ: I know that this is a pattern in our other components, but I am not quite happy with that. It should work without duplicity (and duplicity tends to errors).
> If empty, will default to any exceptions listed in the method argument list.
```suggestion
    @ExceptionHandler
    public ResponseEntity<ErrorResponse> handleValidationException(final MethodArgumentNotValidException ex) {
```

docs-private/Developer-How-To-Start.md
```diff
@@ -32,28 +43,40 @@ liquibase --changelog-file=./docs/db/changelog/changesets/signer-cloud-server/db
...
 
-
-#### PostgreSQL
-
 ```shell
+# PostgreSQL
 liquibase --changeLogFile=./docs/db/changelog/changesets/signer-cloud-server/db.changelog-module.xml --output-file=./docs/sql/postgresql/generated-postgresql-script.sql updateSQL --url=offline:postgresql
 ```
 
+## PowerAuth
```
banterCZ, line 61: Nitpick: Not a Wultra convention (yet), just my obsession with white space. I prefer two empty lines before the header. I consider that well arranged to read in the raw format.

banterCZ (commented):
Great job! Let's improve the return type from the service and tidy up the private documentation a little bit, and we are ready to merge.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
banterCZ: Moreover, here you are losing the cause exception and the whole stacktrace.

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: Fixed

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: You are right. It does some formatting in IntelliJ preview for MD but it is not according to the documentation. I used the recommended syntax.

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: Updated

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: Added the empty line

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: I was a little worried that a level 5 heading might be too deep. I put it back if that’s not an issue.

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: Ok, I removed all these `\`

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: Sorry, wrong copy-paste. Fixed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Ok, I put it into my note as tech debt and we can discuss about it. I feel that rollback for the REST API could be separate task (if we want to do it)

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
michal-rozehnal-w: Right, if we don't want the noise of `errorMessage: null` for success then I also prefer `if`. It is more readable

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Ok, thank you for the inspiration. I created `Try` interface with 2 implementations `Success` and `Error`. Both without any return value since for this endpoint we just return OK/FAIL (and reason in case of error). For next endpoints we can add some parameterized value.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
banterCZ: I think that we do not need the rollback for REST, bwt. it is more Saga pattern than rollback https://learn.microsoft.com/en-us/azure/architecture/patterns/saga

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/Try.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.wultra.signercloud.server.restapi;
```
banterCZ, line 18: Great. We may move it later on. This pattern is more general than REST only.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/ErrorResponse.java
michal-rozehnal-w: I like the idea. Let's have the discussion

docs-private/Developer-How-To-Start.md
```diff
@@ -7,11 +7,28 @@
...
 
+We support:
+- PostgreSQL
+- H2 for test purposes
+
+For local development you can use docker image
+```shell
+docker run --name postgres-liquibase -e POSTGRES_USER=signer_cloud_server -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_DB=signer_cloud_server -p 5432:5432 -d postgres:16
+```
+Don't forget to run the Liquibase script (see the section below); it is not applied automatically.
```
banterCZ, line 28:
```suggestion

Don't forget to run the Liquibase script (see the section below); it is not applied automatically.
```

docs-private/Developer-How-To-Start.md
```diff
@@ -40,20 +57,42 @@ liquibase --changeLogFile=./docs/db/changelog/changesets/signer-cloud-server/db.
...
-        --json '{
-        "end_entity_profile_name": "ExampleEEP",
-        "certificate_authority_name": "ExampleCA",
-        "certificate_request": "MIICh...V8shQ== OR -----BEGIN CERTIFICATE REQUEST-----\nMIICh...V8shQ==\n-----END CERTIFICATE REQUEST-----",
-        "certificate_profile_name": "ENDUSER"
-        }'
+curl --location 'https://smoke-ejbca-dev.wultra.app/ejbca/ejbca-rest-api/v1/certificate/pkcs10enroll' \
+--cert client.crt \
+--key client.key \
+--json '{
```
banterCZ:
```suggestion
     --cert client.crt \
     --key client.key \
     --json '{
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/ValidationExceptionHandler.java
michal-rozehnal-w: Very good observation. I am also agains the duplicity if it is really not needed. Updated

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -0,0 +1,76 @@
...
+    SignerResponse createUpdate(@Valid @RequestBody final CreateUpdateSignerRequest requestBody) {
+        logger.info("action: createUpdateSigner, state: initiated, userId: {}, externalSignerId: {}", requestBody.userId(), requestBody.signerId());
+        final var result = signerService.createUpdateSigner(requestBody);
+
+        if (result.isSuccess()) {
+            logger.info("action: createUpdateSigner, state: succeeded");
+            return new SignerResponse(SignerResponseResult.OK, null);
+        } else {
+            logger.error("action: createUpdateSigner, state: failed, errorMessage: {}", result.getError().getMessage());
+            return new SignerResponse(SignerResponseResult.FAIL, result.getError().getMessage());
```
banterCZ, line 72: I'm afraid that this leads to status code 200, but yes, it is according to the analysis by @tech-kratas

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/Try.java
michal-rozehnal-w: I agree, I also added TODO for parameterized value, so we can add it when we need it.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/Try.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Wrapper for the result of an operation from the service layer.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public sealed interface Try permits Try.TrySuccess, Try.TryError {
+    // TODO: Currently, there is no parameterized value for the return, because so far we don't have any service that requires it.
```
banterCZ: No strong requirement, but I prefer formalized TODO such as `// TODO (racansky, 2025-08-01) some comment`, it helps filtering and in case of reformating keeps the original date.

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: Updated

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: Added the empty line

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/Try.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    static Try error(Throwable t) {
+        return new TryError(t);
+    }
+
+    /**
+     * Represents a successful operation.
+     */
+    final class TrySuccess implements Try {
```
banterCZ: Could be `record`

docs-private/Developer-How-To-Start.md
michal-rozehnal-w: Ok, I added these empty line to all headings

banterCZ (approved):
Thanks for reflecting my comments!

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/Try.java
michal-rozehnal-w: Changed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/Try.java
michal-rozehnal-w: ok, I think it will be beneficial if we use same structure. I added Live template for todo into my IntelliJ

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
michal-rozehnal-w: We no longer have the `ResponseEntity` and agreed we won't use the `ObjectResponse`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: In final version everything is in the controller

PR #41, opened 2025-08-14
Fix #13: Retention/cleanup job

### LINKED ISSUE #13: Retention/cleanup job
### Description

Implement a document cleanup job so the database doesn't grow indefinitely. 

### Acceptance criteria
- document age is tracked and compared to the configuration threshold
- document is removed if the document age exceeds configured threshold

### Technical specification

Document can be in various states (WAITING, REJECTED, SIGNED)... There should be a separate job configuration for each document state.

[Document states](https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/Configuration.md#cloudsigner-server)

[Configuration - CloudSigner Server](https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/Configuration.md#cloudsigner-server)

### QA specification
- test job settings for each state
- check document removal in DB for each state

banterCZ:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/CleanupJob.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private final DocumentService documentService;
+
+    @Scheduled(cron = "${signer-cloud.server.document.cleanup.cron:0 0/10 * * * *}", zone = "UTC")
+    @SchedulerLock(name = "cleanupDocuments")
+    public void cleanupDocuments() {
+        logger.info("action: cleanupDocuments, state: initiated");
+        LockAssert.assertLocked();
+        final var result = documentService.cleanupDocuments();
+        logger.info("action: cleanupDocuments, state: succeeded, {}", result);
```
banterCZ, line 45: Could be for example `action: cleanupDocuments, state: succeeded, CleanupResult[rejectedDocuments=0, signedDocuments=disabled, waitingDocuments=22]`

docs/db/changelog/changesets/signer-cloud-server/1.0.x/db.changelog-version.xml
```diff
@@ -4,8 +4,8 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
                                        https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
 
-    <!-- 1.0.x -->
```
banterCZ, line 7: IMHO, this is superfluous, since we are in the folder `1.0.x`

signer-cloud-server/src/main/resources/application-dev.properties
```diff
@@ -1,5 +1,5 @@
 spring.datasource.url=jdbc:postgresql://localhost:5432/signer_cloud_server
-spring.datasource.username=postgres
+spring.datasource.username=signer_cloud_server
```
banterCZ, line 2: To be the same as described for Liquibase at `Developer-How-To-Start.md`.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -25,6 +25,12 @@ signer-cloud.server.ejbca.rest-client-configuration.key-alias=${EJBCA_KEY_ALIAS:
 signer-cloud.server.ejbca.rest-client-configuration.key-password=${EJBCA_KEY_PASSWORD:}
 signer-cloud.server.ejbca.rest-client-configuration.key-store-bytes=${EJBCA_KEYSTORE_BYTES:}
 
+signer-cloud.server.document.cleanup.cron=2 1 0 * * *
```
banterCZ, line 28: Little randomness. Usually, a lot of stuff executes exactly at 0:00, but since `retention-period` is turned off by default, should we turn the job off, too? cc @tech-kratas

docs/Configuration.md
```diff
@@ -6,12 +6,13 @@ For time-defining values, the following syntax is supported: 1s 1h 1d 1y.
...
-| document.rejected.retentionPeriod | _empty_ | Retention period for rejected documents in days. Empty value means no retention period is used, value `0` means documents will be deleted immediately. |
-| document.signed.retentionPeriod   | _empty_ | Retention period for signed documents in days. Empty value means no retention period is used, value `0` means documents will be deleted immediately.   |
-
+| Property                                              | Default       | Note                                                                                                                                           |
+|:------------------------------------------------------|:--------------|:-----------------------------------------------------------------------------------------------------------------------------------------------|
+| signer-cloud.server.document.waiting.timeout          | `3600s`       | Maximal timeout threshold when is possible sign the document after upload.                                                                     |
```
banterCZ, line 11: 👀 @tech-kratas documentation changes

signer-cloud-server/src/main/resources/com/wultra/signercloud/server/document/DocumentServiceTest.sql
```diff
@@ -0,0 +1,14 @@
...
+    (1, '1'),
+    (2, '2'),
+    (3, '3'),
+    (4, '4');
+
+INSERT INTO sc_document (id, document_content_id, document_id, document_name, external_id, file_name, file_size, hash, signature, signer_id, status, timestamp_created) VALUES
+    (1, 1, '1', 'test', '1', 'test.pdf', 100, '1', '1', 1, 'WAITING', now() - interval '2' day),
+    (2, 2, '2', 'test', '2', 'test.pdf', 100, '1', '1', 1, 'REJECTED', now() - interval '1' hour),
+    (3, 3, '3', 'test', '3', 'test.pdf', 100, '1', '1', 1, 'SIGNED', now() - interval '2' day),
+    (4, 4, '4', 'test', '4', 'test.pdf', 100, '1', '1', 1, 'WAITING', now() - interval '1' hour);
```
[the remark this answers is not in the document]
banterCZ: Well spotted!

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/SchedulerConfig.java
```diff
@@ -0,0 +1,49 @@
...
+
+import javax.sql.DataSource;
+
+/**
+ * Scheduler configuration.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@EnableScheduling
+@EnableSchedulerLock(defaultLockAtLeastFor = "100ms", defaultLockAtMostFor = "10m")
```
michal-rozehnal-w, line 36: Out of curiosity, are these values based on experience with ShedLock? They seem reasonable to me. I was just thinking about making them configurable in the properties file, but then we wouldn’t be able to use the annotation, and it would probably be overengineered. I prefer the annotation if we don't have need to adjust it.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -0,0 +1,91 @@
...
+        if (retentionPeriod != null) {
+            long deletedCount = documentRepository.deleteByStatusAndTimestampCreatedBefore(status, now.minus(retentionPeriod));
+            resultConsumer.accept(String.valueOf(deletedCount));
+        } else {
+            resultConsumer.accept("disabled");
+        }
+    }
+
+    @Builder
+    record CleanupResult(String rejectedDocuments, String signedDocuments, String waitingDocuments) {
```
michal-rozehnal-w, line 89: Why is this record here and not in its own file? It’s definitely not wrong; I’m just asking out of curiosity.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@SpringBootTest(properties = {
+        "signer-cloud.server.document.waiting.retention-period=1d",
+        "signer-cloud.server.document.rejected.retention-period=0",
+        "signer-cloud.server.document.signed.retention-period="
+})
+@ActiveProfiles("test")
+@Transactional
+@Sql
```
michal-rozehnal-w, line 41: I usually set things up directly in the test (via entity and repository classes), which is just my personal preference. But this test with the SQL is very elegant and easy to read—I like it.

michal-rozehnal-w (approved):
It looks good, I just added few out of curiosity comments.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/SchedulerConfig.java
banterCZ: These are only default values if we forgot to specify at the job level (where we do have properties). `LockAtLeastFor` is important when the job is too quick and the lock is released, so more nodes would execute it. This helps us save the database and other resources.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
banterCZ: Not a strong opinion. Just highlighting that the record should be used only in the context of `DocumentService`, not elsewhere.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
banterCZ: I was about to initialize it in Java, but in the end, it was too much code, from my point of view.

Issue #10, opened 2025-07-07
CA Connector


tech-kratas:
### Description

Server needs to connect to the CA to generate a new certificate or revoke existing certificates.

Currently supported platform will be EJBCA.

### Acceptance criteria

- CloudSigner is able to connect to the EJBCA using one of supported auth methods.

### Technical specification

EJBCA auth:
- using generated client management certificate
- access is delegated to the proxy (prefered use case)

See descption of CloudSigner's  "Create New Signer" and "Change Signer Status" API methods for more context.

### QA specification

- Use Create New Signer method to get a new certificate
- Use Change Signer Status method with value REVOKE to revoke certificates

The test needs to be end-to-end (after certificate revocation, the status needs to be checked in both CloudSigner and EJBCA, for example).

banterCZ:
https://docs.keyfactor.com/ejbca/9.0/open-api-specification
/v1/certificate/pkcs10enroll

banterCZ:
https://docs.keyfactor.com/ejbca/latest/ejbca-rest-interface#id-(9.3.3)EJBCARESTInterface-pkcs10enroll.sh

PR #43, opened 2025-08-15
Pass accountBindingId and username to EJBCA

banterCZ:
A follow-up to #10


signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaService.java
```diff
@@ -69,11 +63,32 @@ public X509Certificate enrollCertificate(final String csr) throws IOException, C
...
+
+    /**
+     * Parameter object for enrolling a certificate.
+     *
+     * @param userId user identifier
+     * @param externalSignerId signer identifier
+     * @param csr Certificate Signing Request in PKCS #10 format
+     */
+    @Builder
+    public record ParameterObject(String userId, String externalSignerId, String csr){}
```
banterCZ: 🙁 I do not have a better name. `CertificateRequest` would be a conflict.

michal-rozehnal-w: How about `CertificateEnrollmentRequest` or `CertificateGenerationRequest`?

banterCZ: My concern is that we would have two very similar classes, `CertificateGenerationRequest` versus `CertificateRequest`. 🤔 Maybe I could rename old ones to `*Dto` or `*Rest`, but it does not make me much happier.

banterCZ: 💡I see, I will move the old ones under `EjbcaRestClient`.

michal-rozehnal-w (approved):

PR #49, opened 2025-08-20
Fix #45: Job marking signer as expired

### LINKED ISSUE #45: Job marking signer as expired
### Task specification

Also enqueu callback.

banterCZ:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerConfigurationProperties.java
```diff
@@ -15,23 +15,25 @@
...
+@Getter
+@Setter
+class SignerConfigurationProperties {
 
-    @Test
-    void contextLoads() {
-        // TODO (racansky, 2025-08-01) temporary for context loading until proper tests introduced
+    private Expiration expiration = new Expiration(false, null);
+
+    record Expiration(boolean enabled, String url) {
```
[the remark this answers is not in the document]
banterCZ: Invalid comment, this class is managed by the Spring container, not by programmers.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -61,6 +68,27 @@ Try createUpdateSigner(final CreateUpdateSignerRequest request) {
...
+    long cleanupSigners() {
+        final Instant now = Instant.now();
+        // TODO Lubos limit size
+        final List<Long> ids = signerRepository.markAsExpired(now);
+
+        if (configurationProperties.getExpiration().enabled()) {
+            logger.info("Creating {} expiration callbacks.", ids.size());
+            // TODO Lubos create callback
+            callbackService.save(CallbackEvent.builder()
+                    .build());
```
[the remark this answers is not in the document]
banterCZ: Yes, there is a TODO, will be fixed in consecutive commits.

docs/Integration.md
```diff
@@ -38,6 +38,7 @@ States of the entity (user/device) that can sign documents. Some states can be d
 | BLOCKED | Signer cannot sign documents, certificate renewal is suspended but certificate stays active until its expiration. State can be changed back to ACTIVE. |
 | REMOVED | Signer cannot sign documents, certificate renewal is suspended but certificate stays active until its expiration.                                      |
 | REVOKED | Signer cannot sign documents, certificate renewal is suspended and certificate is immediately revoked.                                                 |
+| EXPIRED | Signer cannot sign documents.                                                                                                                          |
```
banterCZ, line 41: @tech-kratas introducing a new state

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerServiceIntegrationTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @Autowired
+    private JdbcTemplate jdbcTemplate;
+
+    @Test
+    void testCleanupSigners() {
+        final long result = tested.cleanupSigners(1);
+
+        assertEquals(1, result);
+
+        final Map<String, Object> callbackEvent = jdbcTemplate.queryForMap("SELECT * FROM sc_callback_event ORDER BY timestamp_created DESC LIMIT 1");
```
banterCZ, line 41: ℹ️ I do not want to expose `CallbackEventRepository` and have an issue with `BeanPropertyRowMapper`, anyway, for test purposes, `queryForMap` seems acceptable for me.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackEvent.java
```diff
@@ -0,0 +1,62 @@
...
+@ToString
+public class CallbackEvent {
+
+    @Id
+    @Sequence("sc_callback_event_seq")
+    private long id;
+
+    private CallbackType callbackType;
+
+    private String callbackData;
```
banterCZ, line 45: Let's keep it as `String`, we may improve it to `Map` when implementing sending callbacks.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -18,25 +18,35 @@
...
+import java.util.List;
+import java.util.UUID;
 
 /**
  * Service for {@link Signer} operations.
  *
  * @author Michal Rozehnal, michal.rozehnal@wultra.com
  */
 @Service
+@Transactional
```
banterCZ, line 47: Now we have several methods, so I suggest moving the annotation to the class.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -66,6 +77,48 @@ Try createUpdateSigner(final CreateUpdateSignerRequest request) {
...
+        return CallbackEvent.builder()
+                .status(CallbackEventStatus.PENDING)
+                .callbackData(createCallbackData(signer))
+                .callbackType(CallbackType.EXPIRED)
+                .timestampCreated(now)
+                .idempotencyKey(UUID.randomUUID().toString())
+                .build();
+    }
+
+    @SuppressWarnings("java:S5663")
```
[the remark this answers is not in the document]
banterCZ: 👎 This is not related to any logger; I just wanted to avoid escaping in the JSON string.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerRepository.java
```diff
@@ -27,5 +31,54 @@
...
+     * @return List of signers marked as expired.
+     * @implSpec Unfortunately, usage of Common Table Expressions is limited to PostgreSQL only.
+     */
+    default List<Signer> markAsExpired(Instant now, int limit) {
+        final List<Signer> signers = findForExpiration(now, limit);
+        final List<Long> ids = signers.stream()
+                .map(Signer::getId)
+                .toList();
+        markAsExpired(ids, now);
+        return signers;
```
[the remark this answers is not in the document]
banterCZ: This is the DAO layer. It is called in a single transaction from the service.

docs-private/signer_cloud_server_database_schema.puml
```diff
@@ -34,9 +34,24 @@ class sc_document_content <<table>>{
...
+    callback_type varchar(32) NN I (EXPIRED/RENEWED)
+    callback_data text NN
+    status varchar(32) NN I (PROCESSING/PENDING/FAILED/COMPLETED)
+    timestamp_created timestamp(6) NN
+    timestamp_last_call timestamp(6)
+    timestamp_next_call timestamp(6)
+    timestamp_delete_after timestamp(6) I
+    timestamp_rerun_after timestamp(6)
+    attempts integer NN
+    idempotency_key varchar(36) NN
```
michal-rozehnal-w, line 48: I like that you considered idempotency and didn't use an ID for it.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250821-callback-event-table.xml
```diff
@@ -0,0 +1,88 @@
...
+            <column name="callback_type" type="varchar(32)">
+                <constraints nullable="false" />
+            </column>
+            <column name="callback_data" type="text">
+                <constraints nullable="false" />
+            </column>
+            <column name="status" type="varchar(32)" >
+                <constraints nullable="false" />
+            </column>
+            <column name="timestamp_created" type="timestamp(6)" defaultValueComputed="${now}">
```
michal-rozehnal-w, line 26: Does the `defaultValueComputed` work? I remember I tried it for the `sc_signer` table, and it was left as null, which caused an error due to the non-null constraint.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackType.java
```diff
@@ -0,0 +1,29 @@
...
+
+package com.wultra.signercloud.server.callback;
+
+/**
+ * Possible types of a {@link CallbackEvent}.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public enum CallbackType {
+    EXPIRED,
```
michal-rozehnal-w, line 31: Missing javadoc for enum options.

michal-rozehnal-w (commented):
Nice PR, just a few comments

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250821-callback-event-table.xml
banterCZ: Not sure, copied from PAS to make it as similar as possible. Maybe we will extract that as a library.

michal-rozehnal-w: Yep, I did the same (took inspiration from PAS) 😄 . I think the problem is that Spring Data performs a single insert with all properties, so it sends `null` as the timestamp, and the `defaultValueComputed` logic isn’t triggered. Maybe I did something wrong when I tried it for the `Signer`, but it’s worth checking whether the value is set to the default in the database layer.

banterCZ: I set the value programmatically. When I miss it, the insert fails. So it seems the option does not work for Spring Data, but the constraint still checks that, which is good.

michal-rozehnal-w: Another option could be

```
@Column(name = "timestamp_created", updatable = false, insertable = false)
private LocalDateTime timestampCreated;
```

But it is suggestion from Copilot I didn't test.

michal-rozehnal-w (approved):

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250821-callback-event-table.xml
banterCZ: Copilot is lame. This is JPA `@Column`. Spring Data JDBC does not have such options.

michal-rozehnal-w: Right, sorry — I’m still mixing up annotations between JPA and JDBC. In JDBC, we can try the `ReadOnlyProperty` annotation (https://docs.spring.vmware.com/spring-data-jdbc-distribution/docs/3.0.14/reference/html/index.html#jdbc.entity-persistence.read-only-properties). I can test it next week.

PR #50, opened 2025-08-20
Fix #3: REST API Change Signer Status

### LINKED ISSUE #3: REST - Change Signer Status
### Description

Implement PUT method for Change Signer Status.

It's recommended to implement signer management in following order:

- create
- change status / details

### Acceptance criteria

### Technical specification

[Change Signer Status](https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/CloudSigner-API.md#change-signer-status)

[Application States - Signer](https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/Integration.md#signer)

- ACTIVE
- BLOCKED
- REMOVED
- REVOKED - requires synchronously calling CA to revoke all certificates for given entity

### QA specification
- test status change
- test status dependent behavior - change from ACTIVE to BLOCKED and vice versa and REMOVED + REVOKED

michal-rozehnal-w:
It is ready for a full review now


signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
```diff
@@ -75,25 +75,13 @@ public CertificateResponse callPkcs10Enroll(final CertificateRequest request) th
...
 
-        logger.info("Revoking certificate, serial number: {}", parameters.certificateSerialNumber());
-        final ResponseEntity<CertificateRevocationResponse> response = restClient.put(url, new Object(), ParameterizedTypeReference.forType(CertificateRevocationResponse.class));
-
-        Assert.notNull(response.getBody(), "Response body must not be null");
-        logger.info("Got revocation certificate response, revoked: {}", response.getBody().revoked());
-        logger.debug("Got revocation certificate response: {}", response.getBody());
-        return response.getBody();
+        logger.info("Revoking certificates, externalSignerId: {}", externalSignerId);
+        restClient.put(url, null, ParameterizedTypeReference.forType(Void.class));
```
michal-rozehnal-w, line 84: I changed the endpoint because in our last discussion we agreed that we want to revoke all certificates connected to a signer.
Possible improvements:

- The request body is not mandatory, but we can specify a revocation reason if needed.
- There is currently no processing of the response, since it likely does not provide any additional information for us. The important part is the 200 OK status. However, we can add additional logic if required.

Here is the link to the endpoint documentation:
https://docs.keyfactor.com/ejbca/9.1.1/open-api-specification

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaService.java
```diff
@@ -82,6 +82,16 @@ private static X509Certificate convertDerToX509Certificate(final String derCerti
         }
     }
 
+    /**
+     * Revokes all certificates associated with the given {@link com.wultra.signercloud.server.signer.Signer#externalSignerId}.
+     *
+     * @param externalSignerId the external signer ID to identify the signer whose certificates should be revoked
+     * @throws RestClientException if an error occurs during the revocation process
+     */
+    public void revokeCertificates(final String externalSignerId) throws RestClientException {
```
michal-rozehnal-w, line 91: Just wrapping it into service layer so we don't call the `client` directly in `signerService` and we are consistent with certificate enrollment.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -107,4 +115,49 @@ private Signer.SignerBuilder updateSigner(final Signer signer) {
...
+        if (!isTransitionValid) {
+            throw new SignerStatusTransitionException("Invalid status transition from " + oldStatus + " to " + newStatus);
+        }
+
+        if (newStatus == SignerStatus.REVOKED) {
+            ejbcaService.revokeCertificates(externalSignerId);
+        }
+
+        signer = signer.toBuilder()
+                .status(newStatus)
```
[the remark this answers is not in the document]
michal-rozehnal-w: Well spotted. Fixed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -107,4 +115,49 @@ private Signer.SignerBuilder updateSigner(final Signer signer) {
...
+    }
+
+    private void updateStatus(final String externalSignerId, final SignerStatus newStatus) throws RestClientException {
+        var signer = signerRepository.findByExternalSignerId(externalSignerId)
+                .orElseThrow(() -> new SignerNotFoundException("Signer not found for external signer ID: " + externalSignerId));
+
+        final var oldStatus = signer.getStatus();
+
+        if (oldStatus == newStatus) {
+            throw new SignerStatusTransitionException("Signer status is already: " + newStatus);
```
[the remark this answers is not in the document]
michal-rozehnal-w: yep, I was thinking about both approaches and probably just the `return` is better.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -73,4 +70,34 @@ SignerResponse createUpdate(@Valid @RequestBody final CreateUpdateSignerRequest
...
+    SignerResponse updateStatus(@PathVariable final String externalSignerId, @Valid @RequestBody final UpdateSignerStatusRequest requestBody) {
+        logger.info("action: updateSignerStatus, state: initiated, externalSignerId: {}, newStatus: {}", externalSignerId, requestBody.signerStatus());
+        final var result =  signerService.updateStatus(externalSignerId, requestBody);
+
+        if (result.isSuccess()) {
+            logger.info("action: updateSignerStatus, state: succeeded");
+            return new SignerResponse(SignerResponseResult.OK, null);
+        } else {
+            logger.error("action: updateSignerStatus, state: failed, errorMessage: {}", result.getError().getMessage());
+            return new SignerResponse(SignerResponseResult.FAIL, result.getError().getMessage());
```
banterCZ, line 99: Just highlight that we return 200 in case of error. It is according to the documentation, but we will discuss that.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerNotFoundException.java
```diff
@@ -15,30 +15,15 @@
...
-record CertificateRevocationResponse(
-        String issuerDn,
-        String revocationReason,
-        LocalDateTime revocation_date,
-        String serialNumber,
-        String message,
-        boolean revoked,
-        LocalDateTime invalidityDate
-) {
+public class SignerNotFoundException extends RuntimeException {
```
banterCZ, line 25: 🤔 Interesting, that git considers this as _movement_.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -39,6 +42,11 @@
 @AllArgsConstructor
 class SignerService {
 
+    private static final Map<SignerStatus, Set<SignerStatus>> VALID_STATUS_TRANSITIONS = Map.of(
```
banterCZ: It is not critical, but I believe that in this case we should prefer `EnumSet` over `Set`.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -107,4 +115,50 @@ private Signer.SignerBuilder updateSigner(final Signer signer) {
...
+
+        if (oldStatus == newStatus) {
+            return;
+        }
+
+        final var isTransitionValid = VALID_STATUS_TRANSITIONS.getOrDefault(oldStatus, Collections.emptySet())
+                .contains(newStatus);
+
+        if (!isTransitionValid) {
+            throw new SignerStatusTransitionException("Invalid status transition from " + oldStatus + " to " + newStatus);
```
banterCZ: You may use the formatter.
```suggestion
            throw new SignerStatusTransitionException("Invalid status transition from %s to %s".formatted(oldStatus, newStatus);
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -107,4 +115,50 @@ private Signer.SignerBuilder updateSigner(final Signer signer) {
...
+        }
+
+        if (newStatus == SignerStatus.REVOKED) {
+            ejbcaService.revokeCertificates(externalSignerId);
+        }
+
+        signer = signer.toBuilder()
```
banterCZ: I would rather use `final`.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerControllerIntTest.java
```diff
@@ -97,7 +99,7 @@ void testCreateUpdateWhenOperationFailsThenFailResponseIsReturned() throws Excep
         final var request = new CreateUpdateSignerRequest(DUMMY_EXTERNAL_SIGNER_ID, DUMMY_USER_ID, DUMMY_CSR);
 
         // when
-        var mvcResult = mockMvc.perform(post(CREATE_UPDATE_SIGNER_ENDPOINT)
+        final var mvcResult = mockMvc.perform(post(CREATE_UPDATE_SIGNER_ENDPOINT)
```
banterCZ, line 102: In general, there is a danger not escaping the variable, you may use templating something like
```suggestion
        final var mvcResult = mockMvc.perform(post("/api/signers/{id}", DUMMY_EXTERNAL_SIGNER_ID)
```

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerControllerIntTest.java
```diff
@@ -238,30 +240,104 @@ void testCreateUpdateWhenSignerIsUpdatedThenItIsStoredIntoDatabase() throws Exce
...
+        // then
+        final var signer = signerRepository.findAll().iterator().next();
+        assertSigner(signer, DUMMY_TIMESTAMP_CREATED, SignerStatus.BLOCKED);
         assertEquals(Instant.now().toEpochMilli(), signer.getTimestampLastUpdated().toEpochMilli(), MILLISECONDS_DELTA);
     }
 
-    private void assertSigner(Signer signer, Instant expectedTimestampCreated) {
+    private void assertSigner(final Signer signer, final Instant expectedTimestampCreated, final SignerStatus expectedStatus) {
         assertNotEquals(0, signer.getId());
         assertEquals(expectedTimestampCreated.toEpochMilli(), signer.getTimestampCreated().toEpochMilli(), MILLISECONDS_DELTA);
```
banterCZ, line 323: Not bad, I tried `.truncatedTo(ChronoUnit.SECONDS)`.

banterCZ (commented):
A nice draft.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerNotFoundException.java
michal-rozehnal-w: yep, I was surprised too

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: You are right, updated.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Updated

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: I used a new variable `updatedSigner` so the `signer` is not re-assigned.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerControllerIntTest.java
michal-rozehnal-w: Good point, fixed. Just note, I did it for the `put`, in the `post` we don't use variable in path.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerControllerIntTest.java
michal-rozehnal-w: I was also thinking about this, but there’s an edge case when the comparison falls on a second boundary. For example, this will fail:

```
assertEquals(
        LocalDateTime.of(205, 8, 20, 1, 0, 59, 999_999_999).toInstant(ZoneOffset.UTC).truncatedTo(SECONDS),
        LocalDateTime.of(205, 8, 20, 1, 1, 0, 0).toInstant(ZoneOffset.UTC).truncatedTo(SECONDS)
);
```

I admit the probability is very low, but using `assertEquals` with a delta avoids the issue and ensures deterministic tests.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -107,4 +115,50 @@ private Signer.SignerBuilder updateSigner(final Signer signer) {
...
+        try {
+            updateStatus(externalSignerId, request.signerStatus());
+            return Try.success();
+        } catch (SignerNotFoundException | SignerStatusTransitionException | RestClientException e) {
+            return Try.error(e);
+        }
+    }
+
+    private void updateStatus(final String externalSignerId, final SignerStatus newStatus) throws RestClientException {
+        var signer = signerRepository.findByExternalSignerId(externalSignerId)
```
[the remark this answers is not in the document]
michal-rozehnal-w: Updated

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -39,6 +39,11 @@
 @AllArgsConstructor
 class SignerService {
 
+    private static final Map<SignerStatus, EnumSet<SignerStatus>> VALID_STATUS_TRANSITIONS = Map.of(
+            SignerStatus.ACTIVE, EnumSet.of(SignerStatus.BLOCKED, SignerStatus.REMOVED, SignerStatus.REVOKED),
+            SignerStatus.BLOCKED, EnumSet.of(SignerStatus.ACTIVE, SignerStatus.REMOVED, SignerStatus.REVOKED)
+    );
```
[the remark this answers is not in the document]
banterCZ: Well, I think that it is clear enough.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerControllerIntTest.java
```diff
@@ -141,7 +143,7 @@ void testCreateUpdateWhenSignerIsCreatedThenOkResponseIsReturned() throws Except
         final var request = new CreateUpdateSignerRequest(DUMMY_EXTERNAL_SIGNER_ID, DUMMY_USER_ID, DUMMY_CSR);
 
         // when
-        var mvcResult = mockMvc.perform(post(CREATE_UPDATE_SIGNER_ENDPOINT)
+        final var mvcResult = mockMvc.perform(post(CREATE_UPDATE_SIGNER_ENDPOINT)
```
banterCZ, line 146: Objectively, I can't argue with that, but personally, I don't use constants in these cases. I inline the value; it is more convenient for me.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/UpdateSignerStatusRequest.java
```diff
@@ -15,20 +15,14 @@
...
- * @param certificateSerialNumber The hex serial number (without prefix, e.g. '00') of the certificate to be revoked
- * @author Lubos Racansky, lubos.racansky@wultra.com
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
  */
-@Builder
-record CertificateRevocationParameters(
-        String issuerDn,
-        String certificateSerialNumber
-) {
+public record UpdateSignerStatusRequest(@NotNull SignerStatus signerStatus) {
```
banterCZ, line 27: 🤔 We are using the `SignerStatus` for REST API and the database layer as well. Yes, it is simpler, but there are some disadvantages. For example, if I introduce a new value `EXPIRED`, it will be directly exposed to this endpoint. It could be desired or not. Or later refactors may break compatibility.
So far, I am fine with that, but we should realize it at least.

banterCZ (approved):

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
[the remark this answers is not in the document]
michal-rozehnal-w: As I mentioned in one of previous comments, we don't need to process the response. In case of fail there is an `RestClientException`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -28,7 +28,7 @@
 import java.io.IOException;
 import java.security.cert.CertificateException;
 import java.time.Instant;
-import java.util.Base64;
+import java.util.*;
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is default IntelliJ config. I think we don't have strict rules about using the `*`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -49,7 +54,7 @@ class SignerService {
      * (based on {@link CreateUpdateSignerRequest#csr}), and finally stores the signer in the database.
      *
      * @param request the request containing details of signer
-     * @return {@link SignerResponse} indicating the result of the operation.
+     * @return result of operation as {@link Try}
```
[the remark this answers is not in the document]
michal-rozehnal-w: We want to use link to `Try` in javadoc

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerControllerIntTest.java
michal-rozehnal-w: I’m the exact opposite; I like constants 😄. But you’re right that the value of the path isn’t obvious at first glance.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/UpdateSignerStatusRequest.java
michal-rozehnal-w: Good point, you are absolutely right about mixing the public and internal status. I will think about some changes when I will work on another signer endpoint.

michal-rozehnal-w:
~~This PR is ready, but I wasn’t able to test the flow when EJBCA calls for certificates revocation. I’m getting a 403 from the endpoint, which seems to be an issue with the EJBCA settings. I’ll try to resolve the issue with @tech-kratas on Monday and merge it if the testing is successful.~~
Change made. I’ll merge this, and if any issues come up during testing, I’ll fix them as a separate issue. I have another PR based on this branch that’s almost ready, and I don’t want to block it.

PR #54, opened 2025-08-26
Fix #5: REST API Upload document

### LINKED ISSUE #5: REST - Upload Document
### Description

Implement POST method for Upload Document.

It's recommended to implement document management in following order:
- upload
- sign
- download
- reject / delete

### Acceptance criteria

### Technical specification

[Upload Document](https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/CloudSigner-API.md#upload-document)

- upload is realized using multipart/form-data
- we should check maximum file size and mime-type
- we should calculate document hash and return it in the response (signed hash needs to be used in Sign Document method)

### QA specification
- test document upload
- verify returned hash
- test dependent behavior - cleanup job

michal-rozehnal-w:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -85,6 +98,107 @@ private void processRetention(final DocumentStatus status, final Duration retent
...
+            throw new DocumentUploadException("Failed to read file: " + e.getMessage());
+        }
+    }
+
+    private String computeHash(final byte[] content) throws NoSuchAlgorithmException {
+        try {
+            final var digest = MessageDigest.getInstance(HASH_ALGORITHM);
+            final var hashBytes = digest.digest(content);
+            return HexFormat.of().formatHex(hashBytes);
+        } catch (final NoSuchAlgorithmException e) {
```
michal-rozehnal-w, line 196: I presume this exception will never occur, but if it does, we should log it as an `error` because we wouldn’t be able to sign any documents.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -17,53 +17,180 @@
...
+    }
+
+    @Test
+    void testUploadDocumentWhenFileContentCanNotBeReadThenFailResultWithCorrectMessageIsReturned() throws NoSuchAlgorithmException, IOException {
+        // given
+        when(signerRepository.findByExternalSignerId(DUMMY_EXTERNAL_SIGNER_ID)).thenReturn(Optional.of(signer));
+
+        final var file = Mockito.mock(MultipartFile.class);
+        when(file.getContentType()).thenReturn(ContentType.APPLICATION_PDF.getMimeType());
+        when(file.getSize()).thenReturn(Long.valueOf(DUMMY_DOCUMENT_CONTENT.length()));
```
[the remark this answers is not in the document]
michal-rozehnal-w: It’s a personal preference, but I prefer using a method from `Long` rather than casting.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentController.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+    )
+    @PostMapping
+    UploadDocumentResponse upload(
+            @RequestParam("signerId") final String externalSignerId,
+            @RequestParam("externalId") final String externalDocumentId,
+            @RequestParam("name") final String documentName,
+            @RequestParam("content") final MultipartFile file
+    ) throws Throwable {
+        logger.info("action: uploadDocument, state: initiated, externalSignerId: {}, externalDocumentId: {}", externalDocumentId, externalDocumentId);
```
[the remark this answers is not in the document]
michal-rozehnal-w: Well spotted. Fixed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -85,6 +98,107 @@ private void processRetention(final DocumentStatus status, final Duration retent
...
+     * @param documentName name of the document
+     * @param file the PDF document to be stored for signing
+     * @return response as a {@link Try}
+     */
+    Try<UploadDocumentResponse> uploadDocument(
+            final String externalSignerId,
+            final String externalDocumentId,
+            final String documentName,
+            final MultipartFile file
+    ) throws NoSuchAlgorithmException {
```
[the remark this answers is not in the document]
michal-rozehnal-w: It is not wrapped in another exception. It is logged as an error and then re-thrown. It should return a 500 to the user, which I believe is correct. This is purely a server error, and the user cannot do anything to fix it. As I mentioned, this exception will most likely never occur in real life.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/UploadDocumentResponse.java
```diff
@@ -0,0 +1,37 @@
...
+
+import lombok.Builder;
+
+/**
+ * REST API response for {@link Document} upload operation.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@Builder
+public record UploadDocumentResponse(
```
jnpsk: nitpick - this record does not have to be public.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -85,6 +98,107 @@ private void processRetention(final DocumentStatus status, final Duration retent
...
+                .name(documentName)
+                .fileName(fileName)
+                .size(fileSize)
+                .hash(hash)
+                .build();
+    }
+
+    private int getFileSize(final MultipartFile file) {
+        final var size = file.getSize();
+        if (size > Integer.MAX_VALUE) {
```
jnpsk, line 177: Isn't there some stricter upper bound for the file size?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -85,6 +98,107 @@ private void processRetention(final DocumentStatus status, final Duration retent
...
+                .signerId(signer.getExternalSignerId())
+                .externalId(externalDocumentId)
+                .name(documentName)
+                .fileName(fileName)
+                .size(fileSize)
+                .hash(hash)
+                .build();
+    }
+
+    private int getFileSize(final MultipartFile file) {
```
jnpsk: nitpick - this method and the two other bellow can be static

jnpsk (approved):
Nice!
Just added few comments.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/UploadDocumentResponse.java
michal-rozehnal-w: Right, I removed the `public`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Yes, we have a file size limit defined in the properties: `spring.servlet.multipart.max-file-size=50MB`
This serves as protection against `int` casting overflow, because `file.getSize()` returns a `long` but we store the size as an `int`. It is very unlikely that we would ever reach this exception (the file would need to be around 2 GB). However, it is good practice to have this safeguard in place.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Good catch! I made them `static`

PR #61, opened 2025-08-28
Fix #8: REST API Sign document

### LINKED ISSUE #8: REST - Sign Document
### Description

Implement POST method Sign Document.

It's recommended to implement document management in following order:
- upload
- sign
- download
- reject / delete

### Acceptance criteria

- Resulting file is signed and contains user certificate

### Technical specification

[Sign Document](https://github.com/wultra/signer-cloud/blob/issues/1-cloudsigner-documentation/docs/CloudSigner-API.md#sign-document)

Recommended reading:
- ETSI standard [EN 319 142](https://www.etsi.org/deliver/etsi_en/319100_319199/31914201/01.02.01_60/en_31914201v010201p.pdf) (we are targeting PAdES-B-B level)
- [Digital Signatures for PDF Documents by iText](https://itextpdf.com/resources/books/digital-signatures-pdf/thank-you)

Process:
- find upload document based on `documentId`
- verify `signature` with user's public key (stored from enrollment phase)
- if it's correct, assembly document (sign and attach user's certificate)

### QA specification
- try to call endpoint with signed document hash
- check the result - document should contain signature, including certificate with user data

michal-rozehnal-w:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentController.java
```diff
@@ -79,4 +77,37 @@ UploadDocumentResponse upload(
...
+                    )
+            }
+    )
+    @PostMapping("/{documentId}/signature")
+    SignDocumentResponse sign(@PathVariable final String documentId, @Valid @RequestBody final SignDocumentRequest requestBody) throws Throwable {
+        logger.error("action: signDocument, state: initiated, documentId: {}", documentId);
+        final var result = documentService.signDocument(documentId, requestBody);
+
+        if (result.isSuccess()) {
+            logger.error("action: signDocument, state: succeeded");
```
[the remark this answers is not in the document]
michal-rozehnal-w: Sure, I don’t know why I used this error here. Fixed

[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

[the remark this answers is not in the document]
michal-rozehnal-w: Fixed.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+                throw new SignDocumentException("Document signing timeout exceeded");
+            }
+        }
+    }
+
+    private static byte[] verifySignatureAndSignDocument(final String certificateBase64,
+                                                         final String hashBase64,
+                                                         final String hashSignatureBase64,
+                                                         final byte[] documentBytes,
+                                                         final DigestAlgorithm signatureAlgorithm) throws CertificateException {
```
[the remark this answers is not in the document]
michal-rozehnal-w: I think 5 is still fine. Creating a parameter object would be overengineering in this case.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+
+        final var hashBytes = Base64.getDecoder().decode(hashBase64);
+        final var hash = new ToBeSigned(hashBytes);
+
+        final var signatureBytes = Base64.getDecoder().decode(hashSignatureBase64);
+        final var signatureValue = new SignatureValue(signatureParams.getSignatureAlgorithm(), signatureBytes);
+
+        final var isSignatureValid = padesService.isValidSignatureValue(hash, signatureValue, certificateToken);
+        if (!isSignatureValid) {
+            throw new SignDocumentException("Invalid signature");
```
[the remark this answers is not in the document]
michal-rozehnal-w: I think it is ok. This message is returned to the user and we want to keep it short. For troubleshooting we are able to find all these data in DB.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentControllerIntTest.java
```diff
@@ -218,24 +416,51 @@ private void assertUploadResponse(final UploadDocumentResponse response) {
...
+        assertEquals(UPLOADED_FILE_SIZE, document.getFileSize());
         assertEquals(HASH, document.getHash());
         assertEquals(DocumentStatus.WAITING, document.getStatus());
         assertNull(document.getSignature());
 
         final var documentContent = documentContentRepository.findById(document.getDocumentContentId()).orElseThrow();
         assertArrayEquals(Files.readAllBytes(FILE_PATH), documentContent.getContent());
     }
+
+    private void assetSignResponse(final SignDocumentResponse response) {
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+        final var waitingTimeout = configurationProperties.getWaiting().getTimeout();
+        if (waitingTimeout != null) {
+            final var documentSigningDeadline = document.getTimestampCreated().plus(waitingTimeout);
+            if (Instant.now().isAfter(documentSigningDeadline)) {
+                throw new SignDocumentException("Document signing timeout exceeded");
+            }
+        }
+    }
+
+    private static byte[] verifySignatureAndSignDocument(final String certificateBase64,
```
michal-rozehnal-w: I took inspiration from `SignatureTest` which was proof of concept.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+    }
+
+    private static CertificateToken getCertificateToken(final String certificateBase64) throws CertificateException {
+        try {
+            final var certificateBytes = Base64.getDecoder().decode(certificateBase64);
+            final var x509Certificate = (X509Certificate) CertificateFactory.getInstance(CERTIFICATE_TYPE)
+                    .generateCertificate(new ByteArrayInputStream(certificateBytes));
+            return new CertificateToken(x509Certificate);
+        } catch (final CertificateException e) {
+            logger.error("Exception when parsing certificate of type: {}", CERTIFICATE_TYPE, e);
```
michal-rozehnal-w, line 333: We probably want to log `error` here, because in this situation we are not able to work with the certificate at all –> we are not able to sign anything.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+            return stream.readAllBytes();
+        } catch (final IOException e) {
+            throw new SignDocumentException("Failed to read signed document: " + e.getMessage());
+        }
+    }
+
+    private String buildDocumentDownloadUri(final String documentId) {
+        return UriComponentsBuilder.newInstance()
+                .scheme("https")
+                .host(configurationProperties.getDownloadHostname())
```
michal-rozehnal-w: I was also thinking about using `HttpServletRequest`, but it is not reliable when app runs in docker or behind proxy

signer-cloud-server/src/main/resources/application-dev.properties
```diff
@@ -1,6 +1,8 @@
 spring.datasource.url=jdbc:postgresql://localhost:5432/signer_cloud_server
 spring.datasource.username=signer_cloud_server
 
+server.port=8090
```
michal-rozehnal-w, line 4: @korbelm I moved this into `dev` properties. So the prod config will use default port, which is `8080`, and for local development we have `8090`.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -25,7 +25,7 @@ signer-cloud.server.ejbca.rest-client-configuration.use-custom-keystore=true
 signer-cloud.server.ejbca.rest-client-configuration.key-store-password=${EJBCA_KEYSTORE_PASSWORD:}
 signer-cloud.server.ejbca.rest-client-configuration.key-alias=${EJBCA_KEY_ALIAS:}
 signer-cloud.server.ejbca.rest-client-configuration.key-password=${EJBCA_KEY_PASSWORD:}
-signer-cloud.server.ejbca.rest-client-configuration.key-store-bytes=${EJBCA_KEYSTORE_BYTES:}
+signer-cloud.server.ejbca.rest-client-configuration.key-store-bytes=${EJBCA_KEYSTORE_BASE64:}
```
michal-rozehnal-w, line 28: @korbelm and here the variable name change

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentRepositoryTest.java
```diff
@@ -1,153 +0,0 @@
-/*
```
michal-rozehnal-w, line 1: I removed this one because now everything is tested in `DocumentControllerIntTest.java`.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -35,10 +35,14 @@ spring.security.oauth2.resource-server.jwt.audiences=${OAUTH2_AUDIENCES:}
...
 signer-cloud.server.document.cleanup.cron=2 1 0 * * *
 signer-cloud.server.document.waiting.retention-period=
-#signer-cloud.server.document.waiting.timeout=3600 # TODO (racansky, 2025-08-15, #8) verify timeout before signing
+signer-cloud.server.document.waiting.timeout=3600s
 signer-cloud.server.document.rejected.retention-period=
 signer-cloud.server.document.signed.retention-period=
 
+signer-cloud.server.document.download-hostname=${DOCUMENT_DOWNLOAD_HOSTNAME:localhost:8080}
+signer-cloud.server.document.hash-algorithm=SHA-256
+signer-cloud.server.document.signature-algorithm=SHA384
```
jnpsk: Having `hash-algorithm` and `signature-algorithm` parameters both set to a hash function is a bit confusing at the first sight. What does the `signature-algorithm` parameter really set?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+                documentContent.getContent(),
+                configurationProperties.getSignatureAlgorithm());
+
+        final var updatedDocumentContent = documentContent.toBuilder()
+                .content(signedDocumentBytes)
+                .build();
+
+        documentContentRepository.save(updatedDocumentContent);
+
+        final var updatedDocument = document.toBuilder()
```
jnpsk, line 268: Does it work well with Hibernate, as a new instance of the entity is created?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+    Try<SignDocumentResponse> signDocument(final String documentId, final SignDocumentRequest requestBody) {
+        try {
+            final var response = processSignDocument(documentId, requestBody);
+            return Try.success(response);
+        } catch (final CertificateException | DocumentNotFoundException | SignerNotFoundException | SignDocumentException e) {
+            return Try.error(e);
+        }
+    }
+
+    private SignDocumentResponse processSignDocument(final String documentId, final SignDocumentRequest requestBody) throws CertificateException {
```
jnpsk, line 242: This processing IMHO should happen in a transaction

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentController.java
```diff
@@ -79,4 +77,37 @@ UploadDocumentResponse upload(
...
+    @PostMapping("/{documentId}/signature")
+    SignDocumentResponse sign(@PathVariable final String documentId, @Valid @RequestBody final SignDocumentRequest requestBody) throws Throwable {
+        logger.info("action: signDocument, state: initiated, documentId: {}", documentId);
+        final var result = documentService.signDocument(documentId, requestBody);
+
+        if (result.isSuccess()) {
+            logger.info("action: signDocument, state: succeeded");
+            return result.getResponse();
+        } else {
+            logger.info("action: signDocument, state: failed, errorMessage: {}", result.getError().getMessage());
```
jnpsk, line 101: Shouldn't this be a more sever log?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+                .build();
+
+        documentRepository.save(updatedDocument);
+
+        final var downloadUrl = buildDocumentDownloadUri(documentId);
+        return new SignDocumentResponse(documentId, downloadUrl);
+    }
+
+    private void verifyDocumentCanBeSigned(final Signer signer, final Document document) {
+        if (!SignerStatus.ACTIVE.equals(signer.getStatus())) {
```
jnpsk: Nitpick - Using `== operator` or `!= operator` is preferred to `equals` method when comparing enums.

jnpsk (commented):
Nice! I just have a few comments.

signer-cloud-server/src/main/resources/application.properties
michal-rozehnal-w: Good question. They serve different purposes. The `hash-algorithm` is the algorithm used to create a hash of the PDF document (from its content). This hash is returned to the user, and the user signs it using the `signature-algorithm`.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Yes, it works as expected. I tested the full document signing flow, and the record is updated in the database. I’m using Postgres in Docker for local development.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: I agree, and it should already be running in a transaction. We have the `@Transactional` annotation at the class level, and according to the [documentation](https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html#transaction-declarative-annotations-method-visibility), from version 6 package-visible methods are also considered transactional.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentController.java
michal-rozehnal-w: We use `info` in this situation because it is the best match according to [these](https://github.com/wultra/development-internal/wiki/Logging-Guideline#logging-level) rules. The `error` level is probably too strong; perhaps `warn` could be used here too. We will likely improve some global exception handling later.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
jnpsk: ah, ok, I missed that

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Good point! It’s more readable with `==` and `!=`. Fixed.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentController.java
```diff
@@ -79,4 +77,37 @@ UploadDocumentResponse upload(
...
+                    @ApiResponse(
+                            responseCode = "200",
+                            description = "Document signed successfully",
+                            content = @Content(schema = @Schema(implementation = SignDocumentResponse.class))
+                    ),
+                    @ApiResponse(
+                            responseCode = "400",
+                            description = "Invalid document state or invalid signature"
+                    ),
+                    @ApiResponse(
```
banterCZ: Have we already discussed `404 not found` with @tech-kratas ?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
banterCZ: Except for readability, it is null-safe and provides compile-time (static) checking.
https://rules.sonarsource.com/java/RSPEC-4551/

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+            throw new SignDocumentException("Signer is not active. Signer: " + signer.getExternalSignerId());
+        }
+
+        if (!DocumentStatus.WAITING.equals(document.getStatus())) {
```
banterCZ: ditto

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+        params.setSigningCertificate(certificateToken);
+
+        return params;
+    }
+
+    private static byte[] getSignedDocumentBytes(final DSSDocument signedDocument) {
+        try (final var stream = signedDocument.openStream()) {
+            return stream.readAllBytes();
+        } catch (final IOException e) {
+            throw new SignDocumentException("Failed to read signed document: " + e.getMessage());
```
banterCZ, line 353: Please keep the root cause as much as possible.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
banterCZ: Could we resolve headers such as `X-Forwarded-Host`? You may investigate the configuration `server.forward-headers-strategy=framework`.

signer-cloud-server/src/main/resources/application.properties
banterCZ: The explanation is correct. Jan's objection is fair. I do not have a better name yet. Also note that the format is different, `SHA-256` versus `SHA384`

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentRepositoryTest.java
banterCZ: 🤔 It seems suspicious. I tested the method used by the job, not by the controller.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentControllerIntTest.java
```diff
@@ -59,23 +61,42 @@
 class DocumentControllerIntTest {
 
     private static final String UPLOAD_DOCUMENT_ENDPOINT = "/api/documents";
+    private static final String SIGN_DOCUMENT_ENDPOINT = "/api/documents/{documentId}/signature";
+    private static final String CONTENT_TYPE = "application/pdf";
+    private static final String DOCUMENT_NAME_PARAM = "name";
+    private static final String EXTERNAL_DOCUMENT_ID_PARAM = "externalId";
+    private static final Path FILE_PATH = Path.of("./src/test/resources/input.pdf");
```
banterCZ: 🤔 This looks dangerous for CI, since it is using the working directory. Have you considered resources loaded via classpath?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
             throw e;
         }
     }
 
+    /**
+     * Signs the {@link Document} identified by {@code documentId} if it is in the correct state and signature is valid.
+     *
+     * The document is in correct state if:
+     * <ul>
+     *  <li> linked {@link Signer} is in {@link SignerStatus#ACTIVE} status </li>
```
banterCZ, line 222: Nitpick: Why a leading white-space?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+            }
+        }
+    }
+
+    private static byte[] verifySignatureAndSignDocument(final String certificateBase64,
+                                                         final String hashBase64,
+                                                         final String hashSignatureBase64,
+                                                         final byte[] documentBytes,
+                                                         final DigestAlgorithm signatureAlgorithm) throws CertificateException {
+        final var padesService = new PAdESService(new CommonCertificateVerifier());
```
banterCZ: 🤔 Shouldn't this be a registered Spring bean instead?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+    private static PAdESSignatureParameters getSignatureParameters(final CertificateToken certificateToken, final DigestAlgorithm algorithm) {
+        final var params = new PAdESSignatureParameters();
+        params.setDigestAlgorithm(algorithm);
+        params.setSignatureLevel(SignatureLevel.PAdES_BASELINE_B);
+        params.setSigningCertificate(certificateToken);
+
+        return params;
+    }
+
+    private static byte[] getSignedDocumentBytes(final DSSDocument signedDocument) {
```
banterCZ: I would rather avoid `get*` for non-property.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +204,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+            throw e;
+        }
+    }
+
+    private static PAdESSignatureParameters getSignatureParameters(final CertificateToken certificateToken, final DigestAlgorithm algorithm) {
+        final var params = new PAdESSignatureParameters();
+        params.setDigestAlgorithm(algorithm);
+        params.setSignatureLevel(SignatureLevel.PAdES_BASELINE_B);
+        params.setSigningCertificate(certificateToken);
+
```
banterCZ, line 344: @tech-kratas Shouldn't we include the whole certificate chain for later verification instead of just only the certificate?

banterCZ (commented):
Let's fine-tune several details.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Thank you for the hint; I’ll look into it.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Weird, I remember fixing it on Friday, but I didn’t see the change. I probably didn’t commit it at all. It’s really fixed now.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Fixed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: You are right. I added an error log (as we have for the CertificateException). Additionally, we can also add the root exception to `SignDocumentException` (or, more generally, to any of our custom exceptions). What do you think?

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentControllerIntTest.java
michal-rozehnal-w: Good point! It works now, but I used `ClassPathResource`, which is more reliable.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: I just wanted some indentation to make it more readable. I used a tab and IntelliJ converted it to spaces. I can either align it with `<ul>` or add more spaces (2 or 4).

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Good idea! It is in separate config now.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Okay, that’s reasonable. Fixed.

signer-cloud-server/src/main/resources/application.properties
jnpsk: Ok, so it seems those two parameters should not be unified.
Anyway the `signature-algorithm` is not really a signature algorithm here. Can we make it expect a specific signature algorithm from the [list here](https://ec.europa.eu/digital-building-blocks/DSS/webapp-demo/apidocs/eu/europa/esig/dss/enumerations/SignatureAlgorithm.html)? Alternatively we can make it something like `signature.hash-algorithm`

michal-rozehnal-w: I like the idea of naming it `signature-hash-algorithm`.
I also renamed the document hash property to `content-hash-algorithm`, which should be more descriptive.

For both, I used the `DigestAlgorithm` enum as the value. We could also use `SignatureAlgorithm` and then call `getDigestAlgorithm()` to obtain a corresponding `DigestAlgorithm`. The only drawback is that for some values this method returns `null`. That's why I used the `DigestAlgorithm` type in configuration, I think it gives us more flexibility (although it can still throw a `NoSuchAlgorithmException`).
That said, if you would prefer using `SignatureAlgorithm`, let me know and I’ll adjust the code.

banterCZ: 👍 `content-hash-algorithm` sounds good to me.

banterCZ: ~~🤔 But using implicit `getDigestAlgorithm` reduces the complexity of the configuration option.~~

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +203,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+    private static String computeHash(final byte[] content, final DigestAlgorithm hashAlgorithm) throws NoSuchAlgorithmException {
         try {
-            final var digest = MessageDigest.getInstance(HASH_ALGORITHM);
+            final var digest = hashAlgorithm.getMessageDigest();
             final var hashBytes = digest.digest(content);
-            return HexFormat.of().formatHex(hashBytes);
+            return Base64.getEncoder().encodeToString(hashBytes);
         } catch (final NoSuchAlgorithmException e) {
-            logger.error("Hash algorithm not found: {}", HASH_ALGORITHM, e);
+            logger.error("Hash algorithm not found: {}", hashAlgorithm, e);
```
banterCZ, line 212: Doesn't it lead to double logging?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
banterCZ: @jnpsk there is no Hibernate in this project, just JDBC Spring Data.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +203,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+        final var downloadUrl = buildDocumentDownloadUri(documentId);
+        return new SignDocumentResponse(documentId, downloadUrl);
+    }
+
+    private void verifyDocumentCanBeSigned(final Signer signer, final Document document) {
+        if (SignerStatus.ACTIVE != signer.getStatus()) {
+            throw new SignDocumentException("Signer is not active. Signer: " + signer.getExternalSignerId());
+        }
+
+        if (DocumentStatus.WAITING != document.getStatus()) {
```
banterCZ: Personally, I would reorder it, value first, constant second.
```suggestion
        if (document.getStatus() != DocumentStatus.WAITING) {
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -188,17 +203,164 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+                throw new SignDocumentException("Document signing timeout exceeded");
+            }
+        }
+    }
+
+    private byte[] verifySignatureAndSignDocument(final String certificateBase64,
+                                                         final String hashBase64,
+                                                         final String hashSignatureBase64,
+                                                         final byte[] documentBytes,
+                                                         final DigestAlgorithm signatureAlgorithm) throws CertificateException {
```
banterCZ: We do not have conventions for that ye,t and it is difficult to edit IDE formatter, but I would write:

```suggestion
    private byte[] verifySignatureAndSignDocument(
            final String certificateBase64,
            final String hashBase64,
            final String hashSignatureBase64,
            final byte[] documentBytes,
            final DigestAlgorithm signatureAlgorithm) throws CertificateException {
```

banterCZ (commented):
Looks good. I would like to decide on two points:
- Mainly certificate chain.
- Whether the configuration of the host is needed.

signer-cloud-server/src/main/resources/application.properties
jnpsk: yes, now it is more clear

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: It should not (assuming I didn’t overlook anything). When the exception is thrown, it is logged here as an `error` and propagated. After that, there is no catch block for it, and it is not logged elsewhere.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Updated.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Yep, it looks better. Updated.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: I used the `X-Forwarded-*` headers, tested locally, and it works as expected. I like this solution more than having a separate configuration for it. Just one thing to keep in mind: these headers are out of our control and depend on the proxy configuration in front of this app. I mentioned this in the Swagger documentation for `SignDocumentResponse.class`

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -190,7 +438,36 @@ private void assertSuccessResult(final Try<UploadDocumentResponse> result) {
...
+    private static Signer createSigner(final SignerStatus status) {
+        return Signer.builder()
+                .id(SIGNER_ID)
+                .externalSignerId(DUMMY_EXTERNAL_SIGNER_ID)
+                .certificate(CERTIFICATE)
+                .status(status)
+                .build();
+    }
+
+    private void setRequestContext() {
```
banterCZ: I would rather avoid `set*`. What about `create` or `prepare`? Moreover, it could be static, right?

michal-rozehnal-w: The `prepare` name looks good, or `initRequestContext` could also work.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentController.java
michal-rozehnal-w: Not yet; I have started the conversation in the Slack channel.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: I also mentioned this point in the Slack

banterCZ (approved):
Ok, it seems that the certificate chain could be postponed to a separate issue.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: I added TODO into code and create a separate issue for it.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentController.java
michal-rozehnal-w: I updated the code. When resource is not found then HTTP status 400 is returned with `ErrorCode.ERROR_RESOURCE_NOT_FOUND`

PR #66, opened 2025-09-03
Fix #48: Job sending callback

### LINKED ISSUE #48: Job sending callback
### Task specification

_No response_

banterCZ:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -97,17 +93,29 @@ long cleanupSigners(final int limit) {
...
+
+    private static CallbackEvent createCallbackEvent(final Signer signer, final Callback callback, final LocalDateTime now) {
         return CallbackEvent.builder()
                 .status(CallbackEventStatus.PENDING)
                 .callbackData(createCallbackData(signer))
-                .callbackType(CallbackType.EXPIRED)
+                .callbackId(callback.getId())
                 .timestampCreated(now)
                 .idempotencyKey(UUID.randomUUID().toString())
                 .build();
```
jnpsk, line 114: It gets status PENDING, but the timestamp of next call is not set, which seems crucial for the scheduled job to find it.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -44,8 +44,31 @@ signer-cloud.server.document.signature-hash-algorithm=SHA384
 
 signer-cloud.server.signer.expiration.job.cron=3 2 0 * * *
 signer-cloud.server.signer.expiration.job.limit=1000
-signer-cloud.server.signer.expiration.callback.enabled=true
-signer-cloud.server.signer.expiration.callback.url=
+signer-cloud.server.signer.expiration.callback-enabled=true
+
+signer-cloud.server.callback.dispatch-pending-callback-events.job.cron=0 1 * * * *
```
jnpsk: So by default callbacks are dispatched only once an hour?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackService.java
```diff
@@ -40,4 +71,244 @@ public class CallbackService {
...
+        // This is an arbitrary value, representing allowed delay before trying establishing remote connection.
+        final Duration allowedProcessingDelay = Duration.ofSeconds(10);
+        return configuration.getHttpConnectionTimeout()
+                .plus(configuration.getHttpResponseTimeout())
+                .plus(allowedProcessingDelay);
+    }
+
+    private boolean shouldBeSentAtMostOnce(final Callback callback) {
+        return obtainMaxAttempts(callback) == 1;
+    }
```
jnpsk: This was added to PAS mainly for compatibility. The old callback mechanism approach followed "at most once" (fire-and-forget) delivery model. To keep that, we set the default max attempts to 1 and added the explicit condition "shouldBeSentAtMostOnce" to prevent forced reruns, which could otherwise trigger a callback, which was actually delivered, but the response was not received.
When max attempts is 2 or more, it effectively becomes "at least N-times" strategy due to forced reruns.
I mention this so you can reconsider whether this behavior is required in this case as well.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -44,8 +44,31 @@ signer-cloud.server.document.signature-hash-algorithm=SHA384
...
+signer-cloud.server.callback.dispatch-pending-callback-events.job.limit=100
+signer-cloud.server.callback.cleanup-callback-events.job.cron=0 11 * * * *
+signer-cloud.server.callback.rerun-stale-callback-events.job.cron=0 23 * * * *
```
jnpsk: So by default stale callbacks are re-dispatched only once an hour?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackService.java
```diff
@@ -40,4 +71,244 @@ public class CallbackService {
     public void save(final Iterable<CallbackEvent> events) {
         callbackEventRepository.saveAll(events);
     }
```
jnpsk: Do you plan to add a trigger of a callback right at the moment it’s created?

jnpsk (commented):
OK! I am just adding some comments to consider.

signer-cloud-server/src/main/resources/application.properties
banterCZ: Good question, not sure yet, it is to be discussed with @tech-kratas, but in general, certificate validity is in days, so renewal or expiration events do not have to be so granular.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackService.java
banterCZ: Thanks for noticing. In that case, I suggest dropping `shouldBeSentAtMostOnce` to keep it simple. @tech-kratas, any opinion?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackService.java
banterCZ: 🤔 Hmm, you mean `CallbackUrlEventQueueService`, right? https://github.com/wultra/powerauth-server/blob/develop/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/callbacks/CallbackUrlEventQueueService.java
Probably, I simplified it too much, and with this approach, the event would be sent after an hour.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
banterCZ: `timestampNextCall` set.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerServiceIntegrationTest.java
```diff
@@ -1,7 +1,6 @@
 package com.wultra.signercloud.server.signer;
 
 import com.wultra.signercloud.server.callback.CallbackEventStatus;
-import com.wultra.signercloud.server.callback.CallbackType;
 import org.junit.jupiter.api.Test;
```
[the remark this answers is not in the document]
banterCZ: This is our convention.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -18,13 +18,11 @@
 package com.wultra.signercloud.server.signer;
 
 import com.wultra.core.rest.client.base.RestClientException;
-import com.wultra.signercloud.server.callback.CallbackEvent;
-import com.wultra.signercloud.server.callback.CallbackEventStatus;
-import com.wultra.signercloud.server.callback.CallbackService;
-import com.wultra.signercloud.server.callback.CallbackType;
+import com.wultra.signercloud.server.callback.*;
```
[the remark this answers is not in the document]
banterCZ: This is our convention.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackService.java
banterCZ: `CallbackQueueService` added.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -33,10 +31,8 @@
 import java.io.IOException;
 import java.security.cert.CertificateException;
 import java.time.Instant;
-import java.time.LocalDateTime;
 import java.util.*;
```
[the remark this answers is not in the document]
banterCZ: This is our convention.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackEventRepository.java
```diff
@@ -18,12 +18,65 @@
...
+     * @param timestamp the timestamp used to filter callback events with a next call time prior to this value
+     * @param limit the maximum number of callback events to return
+     * @return a list of pending callback events that match the specified criteria
+     * @implSpec {@code FETCH FIRST} is supported by {@code ANSI SQL:2008}.
+     */
+    @Query("""
+            SELECT c FROM sc_callback_event c
+            WHERE c.status = 'PENDING'
+            AND c.timestamp_next_call < :timestamp
+            ORDER BY c.timestamp_next_call DESC
```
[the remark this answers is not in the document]
banterCZ: A good idea.

banterCZ:
Business logic is ready to review. I just have to add some more tests.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerServiceIntegrationTest.java
```diff
@@ -40,9 +39,10 @@ void testCleanupSigners() {
 
         final Map<String, Object> callbackEvent = jdbcTemplate.queryForMap("SELECT * FROM sc_callback_event ORDER BY timestamp_created DESC LIMIT 1");
         assertNotNull(callbackEvent);
-        assertEquals(CallbackType.EXPIRED.toString(), callbackEvent.get("CALLBACK_TYPE"));
-        assertEquals(CallbackEventStatus.PENDING.toString(), callbackEvent.get("STATUS"));
-        assertEquals("{\"externalSignerId\": \"signer1\", \"userId\": \"user1\"}", callbackEvent.get("CALLBACK_DATA"));
+        assertEquals(1L, callbackEvent.get("CALLBACK_ID"));
+        assertEquals(CallbackEventStatus.PROCESSING.toString(), callbackEvent.get("STATUS"));
+        assertEquals("""
```
banterCZ: This payload format is a suggestion. I am open to discussion.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/Callback.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private CallbackType callbackType;
+
+    private String callbackUrl;
+
+    /**
+     * Callback request authentication. May be encrypted, configured by {@link #encryptionMode}.
+     */
+    @Builder.Default
+    private String authentication = "{}";
```
michal-rozehnal-w: Just to clarify: here we can have either JSON (unencrypted option) or Base64 (encrypted option), correct?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackEventResponseHandler.java
```diff
@@ -0,0 +1,159 @@
...
+    private final CallbackEventRepository callbackEventRepository;
+
+    private final LoadingCache<Long, CachedRestClient> callbackRestClientCache;
+
+    /**
+     * Handle a successful Callback Event attempt.
+     *
+     * @param callbackEventData Callback Event successfully delivered.
+     */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
michal-rozehnal-w, line 54: Nitpick: the `@Transactional` annotation can be applied at the class level.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/utils/TransactionUtils.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+/**
+ * Utils class to handle transaction synchronization.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public final class TransactionUtils {
```
michal-rozehnal-w, line 29: This is a good example of a utility class. I remember there is also a Lombok `@UtilityClass` annotation, but I believe it is still experimental and has some limitations. The current implementation is fine.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerServiceIntegrationTest.java
michal-rozehnal-w: It looks good; the most important parts are `externalSignerId` and `callbackType`. However we can modify the payload any time later if we need it.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackConfigurationProperties.java
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Callback configuration properties.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@ConfigurationProperties(prefix = "signer-cloud.server.callback")
+@Getter
+@Setter
+class CallbackConfigurationProperties {
```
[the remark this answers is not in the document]
banterCZ: Spring container reads it even as package private.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/encryption/EncryptionConfigurationProperties.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Encryption configuration properties.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@ConfigurationProperties(prefix = "signer-cloud.server.encryption")
+@Getter
+@Setter
+class EncryptionConfigurationProperties {
```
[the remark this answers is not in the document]
banterCZ: Spring container reads it even as package private.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -87,36 +85,46 @@ long cleanupSigners(final int limit) {
...
+            callbackQueueService.enqueueToDatabase(callbackEventData);
+        }
     }
 
     @SuppressWarnings("java:S5663")
-    private static String createCallbackData(final Signer signer) {
+    private static String createCallbackData(final Signer signer, final CallbackType callbackType) {
         return """
-                {"externalSignerId": "%s", "userId": "%s"}""".formatted(signer.getExternalSignerId(), signer.getUserId());
+                {"externalSignerId": "%s", "userId": "%s", "callbackType": "%s"}""".formatted(signer.getExternalSignerId(), signer.getUserId(), callbackType);
```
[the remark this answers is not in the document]
banterCZ: 🤔 An interesting idea, but we have inputs under control. API is called by another component, not by users.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackEventResponseHandler.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Calculate the back off period for next retry attempt using exponential backoff strategy.
+     *
+     * @param attempts Number of already made attempts.
+     * @param initialBackoff Initial backoff.
+     * @return Duration between last and next attempt.
+     */
+    private static Duration calculateExponentialBackoffPeriod(final int attempts, final Duration initialBackoff, final double multiplier, final Duration maxBackoff) {
```
michal-rozehnal-w, line 121: Out of curiosity, did you consider involve `spring-retry`? We have a rather custom implementation of the job, so we can’t simply use the `@Retryable` annotation. However, it does provide support for backoff calculation. It is just idea, maybe there is not way how we can use it.

michal-rozehnal-w (commented):
It’s a bigger PR, but the business logic looks good to me. Great work! It’s really sophisticated.
Maybe we should provide a brief description of this callback functionality in `docs-private`, since it’s a fairly complex feature. You can always read the logic from the code, but that takes time. I’m thinking of something short for quick orientation in the code.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerServiceIntegrationTest.java
```diff
@@ -40,9 +39,10 @@ void testCleanupSigners() {
...
+                {"externalSignerId": "signer1", "userId": "user1", "callbackType": "EXPIRED"}""", callbackEvent.get("CALLBACK_DATA"));
```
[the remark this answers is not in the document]
banterCZ: Added JSONAssert

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/Callback.java
banterCZ: Exactly.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackEventResponseHandler.java
banterCZ: A nice idea, but I stick with the PAS solution as much as possible. Since Spring 7, it should be part of spring core, see https://spring.io/blog/2025/09/09/core-spring-resilience-features

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackEventResponseHandler.java
banterCZ: Yes, it is possible, but in this case, I want to be explicit, because `REQUIRES_NEW` is quite heavyweight, and when reading the source code, you might be unpleasantly surprised about the class value.

signer-cloud-server/src/main/resources/application.properties
banterCZ: As discussed with @zcgandcomp 

- [x] Run the jobs more often (in minutes)

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/Callback.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Callback event data access object.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@Table("sc_callback")
+@Getter
+@Builder(toBuilder = true)
+@ToString
+public class Callback {
```
banterCZ: Unlike PAS, this configuration is not dynamic.

- [x] Drop the callback table and move the configuration to properties

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackNotificationService.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Callback notification service.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ * @implNote Public class exposed for other modules.
+ */
+@Service
+@AllArgsConstructor
+@Slf4j
+public class CallbackNotificationService {
```
banterCZ: I do not have a better name. We do not have Spring Modulith yet. 🤔 Maybe `CallbackExternalService`?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackRestClient.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Wrapper for the {@link RestClient} to track the creation timestamp and failure statistics of a REST Client instance.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@Builder
+record CallbackRestClient(
+        RestClient restClient,
+        LocalDateTime timestampCreated,
+        AtomicInteger failureCount,
```
banterCZ, line 37: PAS is using a cache for modifying statistics. Here we are changing attributes of a Spring bean, which is why `Atomic` is declared.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackEventResponseHandler.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Handle a successful Callback Event attempt.
+     *
+     * @param callbackEventData Callback Event successfully delivered.
+     */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void handleSuccess(final CallbackEventData callbackEventData) {
+        final CallbackEvent callbackEvent = callbackEventRepository.findById(callbackEventData.id())
+                .orElseThrow(() -> new IllegalStateException("Callback Event was not found in database during its success handling: callbackEventId=" + callbackEventData.id()));
```
michal-rozehnal-w, line 57: At first, I considered creating a custom exception (similar to `DocumentNotFound`). However, on second thought, I think `IllegalStateException` with proper message is fine. This is internal processing, not exposed to the user, and a custom exception probably wouldn’t provide any real benefit.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackNotificationServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Implementation of {@link CallbackNotificationService}.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+@Service
+@AllArgsConstructor
+@Slf4j
+class CallbackNotificationServiceImpl implements CallbackNotificationService {
```
michal-rozehnal-w: It seems that only this single class implements the interface. Can the interface be removed, or do we plan to add more implementations intentionally?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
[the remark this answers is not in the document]
michal-rozehnal-w: Maybe it’s not a bad idea. The user input will probably never be used here, but you never know how the app might evolve in the future.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/callback/CallbackServiceIntegrationTest.java
```diff
@@ -0,0 +1,204 @@
...
+    @AfterAll
+    static void cleanup() throws Exception {
+        mockWebServer.shutdown();
+    }
+
+    @Test
+    void testDispatchInstantCallbackEvent_wrongStatus() {
+        final String idempotencyKey = UUID.randomUUID().toString();
+
+        final long id = callbackEventRepository.save(CallbackEvent.builder()
```
michal-rozehnal-w, line 64: Nitpick: You could move the building of `CallbackEvent `into a separate method since most of the values seem to be the same.

michal-rozehnal-w (commented):
It looks good. Is this PR final, or is there more work to do?

One point to consider (not a must-have): there are a few classes without unit tests:

- CallbackConvertor
- CallbackEventResponseHandler
- CallbackNotificationServiceImpl

Sometimes the logic is very simple, but it still includes some if conditions or try-catch blocks. I believe these cases are already covered in integration tests, so adding a few unit tests would just be a nice-to-have.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackNotificationServiceImpl.java
banterCZ: 🤔 When the name is the same, I get a sonar warning: _Class names should not shadow interfaces or superclasses_ https://rules.sonarsource.com/java/RSPEC-2176/

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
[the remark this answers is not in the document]
banterCZ: I changed it to a record and am using an object mapper.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackNotificationServiceImpl.java
banterCZ: Or what about `CallbackNotificationServiceInternal` or `PrivateCallbackNotificationService`?

banterCZ:
The PR should be final, except for fixing review comments.

banterCZ:
> It looks good. Is this PR final, or is there more work to do?
> 
> One point to consider (not a must-have): there are a few classes without unit tests:
> 
> * CallbackConvertor
> * CallbackEventResponseHandler
> * CallbackNotificationServiceImpl
> 
> Sometimes the logic is very simple, but it still includes some if conditions or try-catch blocks. I believe these cases are already covered in integration tests, so adding a few unit tests would just be a nice-to-have.

Current coverage.
image

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackNotificationServiceImpl.java
michal-rozehnal-w: I see, both are fine, I like the first option with `Internal` little bit more.

michal-rozehnal-w:
> The PR should be final, except for fixing review comments.

Great, I’ll approve it then. There’s nothing major from my review.

michal-rozehnal-w (approved):

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackNotificationServiceImpl.java
banterCZ: Renamed.

PR #75, opened 2025-09-12
Fix #72: Remove /api prefix from endpoints

### LINKED ISSUE #72: Change path of endpoints
### Task specification

Currently we use pattern `/api/documents/`. We agreed the `/api` prefix should be removed.
Fix it in all controllers and don't forget about updating documentation.

michal-rozehnal-w:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/SignDocumentResponse.java
```diff
@@ -32,6 +32,6 @@ record SignDocumentResponse(
         String documentId,
 
         @Schema(description = "URI to access the signed document. It uses X-Forwarded* headers from request for building the URI.",
-                example = "https://signercloud.com:8080/api/v1/documents/3f5e8c7a-2d91-4f9b-bc3e-1a7d2f8e6c42/download")
+                example = "https://signercloud.com:8080/documents/3f5e8c7a-2d91-4f9b-bc3e-1a7d2f8e6c42/download")
```
banterCZ: The change is not from this PR, but are we sure about the `signercloud.com` domain? What about `example.com`, `localhost` or `Wultra`?

banterCZ (approved):

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/SignDocumentResponse.java
michal-rozehnal-w: Good point, I used `example.com`

Issue #76, opened 2025-09-15
Improve exception handling


michal-rozehnal-w:
### Task specification

This is a wrapper issue for several improvements mentioned in the PRs, which can be made in the exception handling area:

1. https://github.com/wultra/signer-cloud/pull/74#discussion_r2347984253
2. https://github.com/wultra/signer-cloud/pull/38#discussion_r2284145788
3. Some endpoints related to Signer returns 200 OK for failed operation. We can return 400 Bad Request with correct error `code` instead.
4. https://github.com/wultra/signer-cloud/issues/89
5. Keep `cause` in our custom exceptions

Additionally we can do this: https://github.com/wultra/signer-cloud/pull/38#discussion_r2284124536

PR #77, opened 2025-09-15
Fix #63: Revoke certificates in EJBCA

### LINKED ISSUE #63: Revoke certificates in EJBCA
### Description

Originally, we wanted to use the EJBCA endpoint to revoke all certificates linked to `externalSignerId`. Unfortunately, the community version does not provide this endpoint. We therefore have 2 options:

1. Call the EJBCA endpoint to get a list of all issued certificates for a given `externalSignerId`, then make a call for each certificate.
2. Create a new table to keep a history of issued certificates. It should contain fields such as `externalSignerId` (with a foreign key to the `sc_signer` table) and `certificateSerialNumber`. We would use this table instead of calling the EJBCA endpoint to retrieve all issued certificates. Still we have to make a one call for each certificate.

So far, we are leaning toward option 2 because we don’t want to be overly dependent on EJBCA. Storing certificate data in `signer-cloud` will also be useful, and we may need this information later when implementing the audit log.

See EJBCA 9.1.1 API [documentation](https://docs.keyfactor.com/ejbca/9.1.1/open-api-specification). Endpoints mentioned in the description above:
`/v1/certificate/{issuer_dn}/{certificate_serial_number}/revoke` - revoke of one certificate
`/v1/endentity/{endentity_name}/revoke` - revoke of all certificates for given `externalSignerId`, not available in community version

### Acceptance criteria

_No response_

### Technical specification

_No response_

### QA specification

_No response_

michal-rozehnal-w:
Ready for review now


docs-private/signer_cloud_server_database_schema.puml
```diff
@@ -52,17 +52,29 @@ class sc_callback_event <<table>> {
     idempotency_key varchar(36) NN
 }
 
+class sc_issued_certificate <<table>> {
+    ID bigint PK
+    timestamp_created timestamp(6) NN
+    signer_id bigint FK I
```
michal-rozehnal-w, line 61: Another option is to use `external_signer_id`, but I think signer `id` is better.

docs-private/signer_cloud_server_database_schema.puml
```diff
@@ -52,17 +52,29 @@ class sc_callback_event <<table>> {
...
+    serial_number varchar(64) NN
+    issuer_dn varchar(512) NN
```
michal-rozehnal-w, line 63: Having the `issuer_dn` for each certificate gives us more flexibility and is safer in case of any changes. However, if we are sure there will always be only one, then this column can be removed and we can keep it in the config.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -159,7 +158,8 @@ private void processCreateUpdateSigner(final CreateUpdateSignerRequest request)
                 .status(SignerStatus.ACTIVE)
                 .build();
 
-        signerRepository.save(signer);
+        final var savedSigner = signerRepository.save(signer);
```
michal-rozehnal-w: There is one possible issue here. If the EJBCA enroll certificate call succeeds, then the certificate exists in EJBCA. However, if this method does not finish successfully, the certificate is not stored in the `sc_issued_certificate` table, and we don’t know that the certificate exists. As a result, we cannot revoke it if requested. That said, we never use the certificate for any signing. The question is whether we should implement rollback logic (e.g., calling the EJBCA endpoint for revocation), or if this is not a problem.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -279,6 +295,42 @@ private void updateStatus(final String externalSignerId, final SignerStatus newS
...
+        try {
+            final var currentCertificate = CertificateUtils.base64ToX509Certificate(signer.getCertificate());
+            revokeCertificates(signer, currentCertificate);
+        } catch (final CertificateException | IOException e) {
+            logger.warn("Error when parsing current certificate", e);
+            throw new CertificateRevocationException("Certificate could not be revoked: " + e.getMessage());
+        }
+    }
+
+    private void revokeCertificates(final Signer signer, final X509Certificate currentCertificate) {
```
michal-rozehnal-w: I would like to keep this logic as simple as possible, but there are a few things to consider

1. The current call is synchronous, but if there are many certificates to revoke, it may take some time. We could make it asynchronous; however, I am not sure whether the ‘call and forget’ approach is acceptable here, or if we would need an endpoint to check the status of this asynchronous operation.
2. Currently, the exception is thrown only when revocation of the current certificate fails. I think revoking other certificates is more of a cleanup task. But please correct me if I’m wrong and the exception should be returned in the case of any certificate failure.
3. We revoke certificates for a given signerId that are not already expired. I wanted to avoid having a status column in the `sc_issued_certificate` table, since we would need to keep the status up to date. However, we can still add it if it brings any benefit.

docs-private/signer_cloud_server_database_schema.puml
banterCZ: Sure, we should use FK pointing to PK.

docs-private/signer_cloud_server_database_schema.puml
```diff
@@ -52,17 +52,29 @@ class sc_callback_event <<table>> {
     idempotency_key varchar(36) NN
 }
 
+class sc_issued_certificate <<table>> {
```
banterCZ: 🤔 It is not the certificate, is it? Only metadata.

docs-private/signer_cloud_server_database_schema.puml
```diff
@@ -52,17 +52,29 @@ class sc_callback_event <<table>> {
...
+    ID bigint PK
+    timestamp_created timestamp(6) NN
```
banterCZ, line 60: Let's distinguish what the identifier is.

```suggestion
    ID bigint PK
    --
    timestamp_created timestamp(6) NN
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
```diff
@@ -75,12 +74,15 @@ public CertificateResponse callPkcs10Enroll(final CertificateRequest request) th
         return response.getBody();
     }
 
-    void revokeCertificates(final String externalSignerId) throws RestClientException {
-        final String url = UriComponentsBuilder.fromPath("/v1/endentity/{external_signer_id}/revoke")
-                .buildAndExpand(externalSignerId)
+    void revokeCertificate(final EjbcaService.RevokeCertificateRequest request) throws RestClientException {
```
banterCZ, line 97: JavaDoc could be handy.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
```diff
@@ -75,12 +74,15 @@ public CertificateResponse callPkcs10Enroll(final CertificateRequest request) th
...
+        final var serialNumber = request.serialNumber();
+        final var issuerDN = request.issuerDN();
+
+        final String url = UriComponentsBuilder.fromPath("/v1/certificate/{issuer_dn}/{certificate_serial_number}/revoke")
+                .buildAndExpand(issuerDN, serialNumber)
                 .toUriString();
 
-        logger.info("Revoking certificates, externalSignerId: {}", externalSignerId);
+        logger.info("Revoking certificate, serialNumber: {}, issuerDN: {}", request.serialNumber(), request.issuerDN());
```
banterCZ: Should we also log after the call response?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationException.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package com.wultra.signercloud.server.signer;
+
+/**
+ * Exception thrown when certificate revocation fails.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public class CertificateRevocationException extends RuntimeException {
+    public CertificateRevocationException(final String message) {
```
banterCZ: I would rather have a constructor with `cause` as well to keep context.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificate.java
```diff
@@ -0,0 +1,55 @@
...
+public class IssuedCertificate {
+
+    @Id
+    @Sequence("sc_issued_certificate_seq")
+    private long id;
+
+    /**
+     * {@link Signer#id}
+     */
+    private long signerId;
```
banterCZ: We should try `AggregateReference`, but let's do it in a separate PR.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificateRepository.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Repository for accessing a {@link IssuedCertificate}.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+interface IssuedCertificateRepository extends CrudRepository<IssuedCertificate, Long> {
+
+    @Query("""
+        SELECT * FROM sc_issued_certificate
+        WHERE signer_id = :signerId AND timestamp_certificate_expiration < :now
```
banterCZ: No strong opinion, but we may use `NOW()` and avoid passing the date from the Java code.
```suggestion
        WHERE signer_id = :signerId AND timestamp_certificate_expiration < NOW()
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
banterCZ: Not sure, let's discuss at the Thursday meeting.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -234,6 +234,22 @@ private Signer.SignerBuilder updateSigner(final Signer signer) {
                 .timestampLastUpdated(Instant.now());
     }
 
+    private void saveIssuedCertificate(final long signerId, final X509Certificate x509Certificate) {
```
banterCZ, line 309: We should merge this with renewal functionality #78

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -279,6 +295,42 @@ private void updateStatus(final String externalSignerId, final SignerStatus newS
         signerRepository.save(updatedSigner);
     }
 
+    private void revokeCertificates(final Signer signer) {
+        try {
+            final var currentCertificate = CertificateUtils.base64ToX509Certificate(signer.getCertificate());
```
banterCZ: Instead of utils, I have moved this method directly to the Signer, see #78

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -279,6 +295,42 @@ private void updateStatus(final String externalSignerId, final SignerStatus newS
...
+            throw new CertificateRevocationException("Certificate could not be revoked: " + e.getMessage());
+        }
+    }
+
+    private void revokeCertificates(final Signer signer, final X509Certificate currentCertificate) {
+        final var certificatesToRevoke = issuedCertificateRepository.findForRevocation(signer.getId(), Instant.now());
+
+        final var certificatesCount = certificatesToRevoke.size();
+        var counter = 1;
+        for (final var certificate : certificatesToRevoke) {
```
banterCZ: Isn't here the old school for statement more elegant? `for (int i = 0; i < list.size(); i++) {`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -279,6 +295,42 @@ private void updateStatus(final String externalSignerId, final SignerStatus newS
...
+
+            try {
+                ejbcaService.revokeCertificate(request);
+                logger.info("Certificate revoked");
+            } catch (final RestClientException e) {
+                logger.warn("Exception when revoking certificate", e);
+
+                if (Objects.equals(certificate.getSerialNumber(), currentCertificate.getSerialNumber().toString()) &&
+                        Objects.equals(certificate.getIssuerDn(), currentCertificate.getIssuerX500Principal().getName())) {
+                    throw new CertificateRevocationException("Certificate could not be revoked because of EJBCA client error: " + e.getMessage());
```
banterCZ: Here I would like to keep the `cause`.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificateRepository.java
```diff
@@ -0,0 +1,38 @@
...
+        """)
+    List<IssuedCertificate> findForRevocation(final long signerId, final Instant now);
```
banterCZ: Are we sure about this logic? Shouldn't we filter status and signerId?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: As we discussed, it is not an issue since such a certificate will never be used in any document.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificateRepository.java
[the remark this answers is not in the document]
michal-rozehnal-w: RIght, fixed

docs-private/signer_cloud_server_database_schema.puml
michal-rozehnal-w: Yes, good point. I renamed it to `sc_issued_certificate_metadata`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
michal-rozehnal-w: Yes, I moved this method into separate service and javadoc is there.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
michal-rozehnal-w: Good idea, added.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationException.java
michal-rozehnal-w: Added for this exception. In https://github.com/wultra/signer-cloud/issues/76 it will be added to all our custom exception.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificate.java
michal-rozehnal-w: It uses the `AggregateReference` now.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificateRepository.java
michal-rozehnal-w: I used the `NOW()` in order to be consistent

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Thank you for reminding. The renew logic also creates the issued certificate metadata record now.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: Probably yes, since we need the index. It uses for loop now.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
michal-rozehnal-w: I removed this code, but we log a cause in case of EJBCA error response.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/CertificateRevocationServiceTest.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private void assertSavedIssuedCertificateMetadata(final IssuedCertificateMetadata issuedCertificateMetadata) {
+        assertEquals(ID, issuedCertificateMetadata.getId());
+        assertEquals(SIGNER_ID, issuedCertificateMetadata.getSigner().getId());
+        assertEquals(TIMESTAMP_CREATED.toEpochMilli(), issuedCertificateMetadata.getTimestampCreated().toEpochMilli(), MILLISECONDS_DELTA);
+        assertEquals(Instant.now().toEpochMilli(), issuedCertificateMetadata.getTimestampLastUpdated().toEpochMilli(), MILLISECONDS_DELTA);
+        assertEquals(SERIAL_NUMBER, issuedCertificateMetadata.getSerialNumber());
+        assertEquals(ISSUER_DN, issuedCertificateMetadata.getIssuerDn());
+        assertEquals(ISSUER_DN, issuedCertificateMetadata.getIssuerDn());
```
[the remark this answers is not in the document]
michal-rozehnal-w: Removed the second one.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
```diff
@@ -48,11 +50,20 @@ class EjbcaRestClient {
...
     public EjbcaRestClient(final EjbcaConfigurationProperties properties) throws RestClientException {
         validateConfiguration(properties.getRestClientConfiguration());
         this.restClient = new DefaultRestClient(properties.getRestClientConfiguration());
     }
 
+    /**
+     * Constructor for unit testing purposes.
+     * @param restClient Rest client to be used.
+     */
+    EjbcaRestClient(final RestClient restClient) {
```
michal-rozehnal-w, line 63: This constructor is only for tests (mocking the `restClient`). I am open to other ideas. I didn’t want to use reflection or access a private field with Mockito.
The unit test is quite simple—it just verifies that the client is called with the correct parameters. However, I spent some time debugging and making sure the call is correct, so I think it is good to have such a test.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
```diff
@@ -77,13 +88,25 @@ public CertificateResponse callPkcs10Enroll(final CertificateRequest request) th
...
+        final var url = UriComponentsBuilder.fromPath("/v1/certificate/{issuerDN}/{serialNumber}/revoke")
+                        .buildAndExpand(issuerDN, serialNumber)
+                        .toUriString();
+
+        final var params = MultiValueMap.fromSingleValue(Map.of("reason", "UNSPECIFIED"));
 
-        logger.info("Revoking certificates, externalSignerId: {}", externalSignerId);
-        restClient.put(url, null, ParameterizedTypeReference.forType(Void.class));
+        logger.info("Revoking certificate, serialNumberHex: {}, issuerDN: {}", request.serialNumberHex(), request.issuerDN());
+        restClient.put(url, null, params, null, ParameterizedTypeReference.forType(Void.class));
```
michal-rozehnal-w, line 109: I used `Void` since we don’t need to process the response—we already know most of the values. Example of response:
```
{
  "issuer_dn": "CN=ExampleCA",
  "revocation_reason": "KEY_COMPROMISE",
  "revocation_date": "1970-01-01T00:00:00Z",
  "serial_number": "1234567890ABCDEF",
  "message": "Successfully revoked",
  "revoked": true,
  "invalidity_date": "1970-01-01T00:00:00Z"
}
```

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationService.java
```diff
@@ -0,0 +1,86 @@
...
+    private final EjbcaService ejbcaService;
+
+    /**
+     * Revokes single certificate in EJBCA and updates {@link IssuedCertificateMetadata#status} to {@link IssuedCertificateStatus#REVOKED}.
+     * This method always runs in a new transaction.
+     *
+     * @param certificateMetadata Certificate metadata to be revoked.
+     */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    void revokeCertificate(final IssuedCertificateMetadata certificateMetadata) {
```
michal-rozehnal-w: I used a new transaction since the revocation of each certificate is an isolated operation, not related to the others.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationService.java
```diff
@@ -0,0 +1,86 @@
...
+                    certificateMetadata.getIssuerDn());
+
+            ejbcaService.revokeCertificate(request);
+
+            setRevokedStatus(certificateMetadata);
+            logger.info("Certificate successfully revoked");
+        } catch (final RestClientException e) {
+            logger.warn("Exception when revoking certificate: {}", e.getResponse(), e);
+
+            if (e.getStatusCode() == HttpStatusCode.valueOf(409) && e.getResponse().contains("has previously been revoked")) {
```
michal-rozehnal-w: The revoke endpoint is not an idempotent action. The first call is successful, but any subsequent call returns a 409. According to the documentation: 
_Only KEY_COMPROMISE is allowed as a new revocation reason if the revocation reason is to be changed._ 
Therefore, when we receive a 409 with a message indicating that the certificate is already revoked, we treat it as a successful operation and set the status to `REVOKED` in our database.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificateStatus.java
```diff
@@ -0,0 +1,35 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.wultra.signercloud.server.signer;
+
+/**
+ * Enumeration representing status of issued certificate.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public enum IssuedCertificateStatus {
```
michal-rozehnal-w, line 25: I wanted to keep these statuses as simple as possible and avoid any extra jobs. So there is ISSUED, indicating that a certificate was enrolled, and that’s all. The second status is REVOKED, indicating that we performed the revoke operation.
Another option would be use some flag column instead of status, for example `revoked` with `true/false`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerService.java
```diff
@@ -333,6 +356,19 @@ private void updateStatus(final String externalSignerId, final SignerStatus newS
...
+    private void revokeCertificates(final Signer signer) {
+        final var signerId = signer.getId();
+
+        final var certificatesMetadata = issuedCertificateMetadataRepository.findForRevocation(signerId);
+        final var certificatesToRevokeCount = certificatesMetadata.size();
+
+        for (var i = 0; i < certificatesToRevokeCount; i++) {
+            logger.info("Revoking certificate {}/{} in EJBCA", i + 1, certificatesToRevokeCount);
+            final var certificateMetadata = certificatesMetadata.get(i);
+            certificateRevocationService.revokeCertificate(certificateMetadata);
```
michal-rozehnal-w: I simplified it, and we set the status of the signer to `REVOKED` only if the REST API calls for all issued certificates are successful, as we discussed on Wednesday.
Since each call has its own transaction, this endpoint/method can be called repeatedly, and certificates that were successfully processed in previous calls are not processed again.

docs-private/signer_cloud_server_database_schema.puml
banterCZ: It is up to @tech-kratas

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
banterCZ: 🤔 Yes, sometimes cripling production code for tests is needed, but better to avoid as much as possible. I do not have exact instructions yet, but you may use an integration test with mock http server or define a RestClient spring bean (in some Spring configuration class) distinguished by a qualifier. Then it is clear for production code, and it is possible to mock in tests. Moreover, `EjbcaRestClient` would not be responsible for the configuration of RestClient anymore, which is good.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationService.java
```diff
@@ -0,0 +1,86 @@
...
+            final var serialNumberHex = new BigInteger(serialNumber).toString(16);
+
+            final var request = new EjbcaService.RevokeCertificateRequest(
+                    serialNumberHex,
+                    certificateMetadata.getIssuerDn());
+
+            ejbcaService.revokeCertificate(request);
+
+            setRevokedStatus(certificateMetadata);
+            logger.info("Certificate successfully revoked");
```
banterCZ: I know, we should see traceId in the log, but may we add the serial number to the context?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationService.java
banterCZ: Isn't it possible to use the constant `HttpStatus#CONFLICT` instead of `valueOf`?
A matching exact error message is quite fragile; it may change quite easily without being noticed.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificateMetadataRepository.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Repository for accessing a {@link IssuedCertificateMetadata}.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+interface IssuedCertificateMetadataRepository extends CrudRepository<IssuedCertificateMetadata, Long> {
+
+    @Query("""
+        SELECT * FROM sc_issued_certificate_metadata
+        WHERE signer_id = :signerId AND timestamp_certificate_expiration > NOW() AND status != 'REVOKED'
```
banterCZ, line 34: 👍 I have checked that indexes are created for both columns.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificateStatus.java
banterCZ: I like the enum; it is more flexible.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/IssuedCertificateMetadataRepositoryIntTest.java
```diff
@@ -0,0 +1,74 @@
...
+    @Test
+    void testFindForRevocation() {
+        // given
+        // - see the SQL script
+
+        // when
+        final var result = repository.findForRevocation(SIGNER_ID);
+
+        // then
+        final var expectedIds = List.of(4L)
```
banterCZ: Nitpick: Interesting, why not `new Long[]{4L};`? But still, this is valid.

banterCZ (approved):
Please try to reconsider the testing of EjbcaRestClient; otherwise good.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/ejbca/EjbcaRestClient.java
michal-rozehnal-w: Thank you for the ideas. I’ll keep them in my notes, and we can improve this later as technical debt, since this PR is already quite large and the testing constructor is not visible outside the `ejbca` package.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationService.java
michal-rozehnal-w: Good idea, I added both `issuerDn` and `serialNumber`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationService.java
michal-rozehnal-w: Right, I used the enum value.
I agree with you that checking the message text is not ideal. On the other hand, I didn’t find a better way—checking just the HTTP status wouldn’t be enough, and unfortunately, we don’t receive more information. In the worst case, if the message is changed and condition not matched, it will be considered as error and revoke operation won't be successful.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/IssuedCertificateMetadataRepository.java
michal-rozehnal-w: Thank you for checking these.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/IssuedCertificateMetadataRepositoryIntTest.java
michal-rozehnal-w: Yep, I just like the `List.of`, but I like your suggestion since it is more minimalistic. Updated

docs/Signer-Cloud-API.md
```diff
@@ -112,16 +112,18 @@ Change the status of an existing signer (e.g., activate, deactivate, suspend) id
...
+| `signerStatus`     | `String` | Select new signer status. ENUM: ACTIVE, BLOCKED, REMOVED, REVOKED, EXPIRED                                                                                                                                                                                                                                                                                                                 |
+| `revocationReason` | `String` | Optional parameter, used only if `signerStatus` is set to REVOKED. It specifies the reason for revocation, which is passed to EJBCA. If not provided, the default value UNSPECIFIED is used. ENUM: NOT_REVOKED, UNSPECIFIED ,KEY_COMPROMISE, CA_COMPROMISE, AFFILIATION_CHANGED, SUPERSEDED, CESSATION_OF_OPERATION, CERTIFICATE_HOLD, REMOVE_FROM_CRL, PRIVILEGES_WITHDRAWN, AA_COMPROMISE |
```
banterCZ: I suggest using backticks for enum values to make it look like code.

michal-rozehnal-w: I added them to all enum values which are mentioned in this documentation

PR #87, opened 2025-09-19
Fix #64: Add certificate chain into signed document

### LINKED ISSUE #64: Add certificate chain into signed document
### Task specification

Add certificate chain param into [DocumentService](https://github.com/wultra/signer-cloud/blob/develop/signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java#L343). ~~It should be configurable and included by default~~. The chain will be stored for each certificate in database.

michal-rozehnal-w:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/utils/CertificateUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Converts a Base64-encoded X.509 certificate in DER format to an {@link X509Certificate} object.
+     *
+     * @param certificateBase64 the Base64-encoded X.509 certificate string in DER format
+     * @return the X509Certificate object
+     * @throws CertificateException in case of an error during certificate parsing
+     */
+    public static X509Certificate base64ToX509Certificate(final String certificateBase64) throws CertificateException {
```
michal-rozehnal-w, line 49: I moved it out of the `Signer` class because it is used across multiple components. Please let me know if you would prefer a different approach. I placed it in the Utility class since the code is small, not enough for its own service.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentControllerIntTest.java
```diff
@@ -763,4 +792,32 @@ private void assertRejectedDocument(final long documentId) {
...
+        validator.setCertificateVerifier(new CommonCertificateVerifier());
+
+        final var report = validator.validateDocument().getSimpleReport();
+        assertEquals(1, report.getSignaturesCount(), "There is not exactly one signature in document");
+
+        final var signatureId = report.getFirstSignatureId();
+        final var chain = validator.getSignatureById(signatureId).getCertificates();
+        final var certificateChainBase64 = chain.stream()
+                .map(certificateToken -> {
+                    try {
```
michal-rozehnal-w, line 813: Not pretty, but since there is a checked exception, there must be a catch.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/utils/CertificateUtils.java
```diff
@@ -0,0 +1,56 @@
...
+        final var certificateBytes = Base64.getDecoder().decode(certificateBase64);
+        final Certificate result = CertificateFactory.getInstance(CERTIFICATE_TYPE)
+                .generateCertificate(new ByteArrayInputStream(certificateBytes));
+        Assert.isInstanceOf(X509Certificate.class, result, "Certificate is must be of type X509Certificate");
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/utils/CertificateUtils.java
banterCZ: Ok, why not, when I have been changing the signer, I did not yet have all the information.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -41,6 +41,7 @@ signer-cloud.server.document.signed.retention-period=
 
 signer-cloud.server.document.content-hash-algorithm=SHA256
 signer-cloud.server.document.signature-hash-algorithm=SHA384
+signer-cloud.server.document.certificate-chain=
```
banterCZ: Should it be taken from EJBCA?
Anyway, missing in `Configuration.md`

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentControllerIntTest.java
banterCZ: Since it is „only“ in test, have you considered https://projectlombok.org/features/SneakyThrows or extracting it to a method?

banterCZ (approved):
The code looks good, just not sure about the source of the certificate chain.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/Signer.java
```diff
@@ -66,18 +67,23 @@ public class Signer {
 
     private SignerStatus status;
 
+    private String certificateChain;
```
michal-rozehnal-w, line 70: Ideally, this should be a `List<String>`, where each item is a Base64-encoded DER certificate. However, Spring does not provide a default converter for storing collections. I considered creating a custom converter (simply joining items with a `,` separator), but with JDBC such a converter would apply to all `List<String>` fields, which we don’t want.

So, I added a builder method and a getter to work with `List<String>`. However, this is not perfect, since it’s still possible to use `certificateChain(String)` in the builder alongside `certificateChainFromList(List<String>)`.

I would welcome any ideas on the best approach to expose working with a List while keeping the serialization private within the class.

docs-private/signer_cloud_server_database_schema.puml
```diff
@@ -12,6 +12,7 @@ class sc_signer <<table>>{
     certificate CLOB
     timestamp_certificate_expiration timestamp(6) I
     status varchar(50) NN I (ACTIVE/BLOCKED/REMOVED/REVOKED/EXPIRED)
+    certificate_chain
```
michal-rozehnal-w, line 15: I let it nullable in order to be consistent with the `certificate` column. But I am thinking that both `certificate` and `certificate_chain` can have not null constraint

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentControllerIntTest.java
michal-rozehnal-w: `@SneakyThrows` sounds good for tests, but I would still need to create a separate method for the encoding. I will probably go with a `try-catch` in the `.map` body. I don't like multiline map body, but in this case it is acceptable

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/Signer.java
```diff
@@ -66,18 +67,23 @@ public class Signer {
...
-        return (X509Certificate) result;
+        return CertificateUtils.base64ToX509Certificate(certificate);
+    }
+
+    public List<String> getCertificateChain() {
+        return Optional.of(certificateChain)
+                .map(i -> i.split(CERTIFICATES_SEPARATOR))
+                .map(Arrays::asList)
+                .orElse(Collections.emptyList());
     }
```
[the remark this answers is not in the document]
michal-rozehnal-w: Well spotted. Fixed

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250919-certificate-chain.xml
```diff
@@ -0,0 +1,20 @@
...
+    <changeSet id="1" logicalFilePath="signer-cloud-server/1.0.x/20250919-certificate-chain.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <tableExists tableName="sc_signer"/>
+                <not>
+                    <columnExists tableName="sc_signer" columnName="certificate_chain"/>
+                </not>
+            </and>
+        </preConditions>
+        <comment>Create a new index column certificate_chain</comment>
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

signer-cloud-server/src/main/resources/application.properties
michal-rozehnal-w: Yes, with the latest implementation a chain is stored into DB for with each certificate and this property no longer exist.

docs-private/signer_cloud_server_database_schema.puml
banterCZ: 🤔 At first glance, I think that at least `certificate` should be NOT NULL, but we may fix it in a separate PR.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250919-certificate-chain.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <changeSet id="1" logicalFilePath="signer-cloud-server/1.0.x/20250919-certificate-chain.xml" author="Michal Rozehnal">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <tableExists tableName="sc_signer"/>
```
banterCZ: Not bad, do we have to write a precondition even for the table, if we have the column check?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -324,12 +327,42 @@ private byte[] verifySignatureAndSignDocument(
...
+
+            for (final var base64Certificate : signer.getCertificateChain()) {
+                final var x509Certificate = CertificateUtils.base64ToX509Certificate(base64Certificate);
+                chain.add(new CertificateToken(x509Certificate));
+            }
+
+            return chain;
+        } catch (final CertificateException e) {
+            logger.warn("Exception when processing certificate chain", e);
+            throw new SignDocumentException("Exception when processing certificate chain: " + e.getMessage());
```
banterCZ, line 352: Have you considered passing the `cause`?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/Signer.java
banterCZ: 🤔 It is a sad story. I have read about making a wrapper for `List<String>`, something like `CertificateChain` class.

banterCZ (approved):

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: Yes, I added it into https://github.com/wultra/signer-cloud/issues/76, so we are sure it is not forgotten and it is done for all our custom exceptions. It will be part of PR for the issue.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250919-certificate-chain.xml
michal-rozehnal-w: Right, it is not really needed. I removed the `tableExists`.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/Signer.java
michal-rozehnal-w: Ok, we can fix it as tech debt later. I need to think more about it and find the best minimalistic solution.
Copilot also suggested the wrapper class as possible solution, but I don't like much.

docs-private/signer_cloud_server_database_schema.puml
michal-rozehnal-w: I agree we should do it as separate PR, it could require fix of a few tests since the certificate is not always set.

PR #96, opened 2025-09-25
Fix #76: Improve exception handling

### LINKED ISSUE #76: Improve exception handling

michal-rozehnal-w:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/RestExceptionHandler.java
```diff
@@ -74,50 +109,107 @@ public ResponseEntity<ErrorResponse> handleSignerNotFoundException(final SignerN
...
-     * Handler for {@link RejectDocumentException} producing {@link HttpStatus#BAD_REQUEST} response.
+     * Handler for generic {@link RuntimeException} producing {@link HttpStatus#BAD_REQUEST} response.
+     * This is a fallback handler for all unhandled runtime exceptions.
      *
      * @param ex the exception
      * @return response as {@link ResponseEntity}
      */
     @ExceptionHandler
-    public ResponseEntity<ErrorResponse> handleRejectDocumentException(final RejectDocumentException ex) {
+    public ResponseEntity<ErrorResponse> handleRuntimeException(final RuntimeException ex) {
```
michal-rozehnal-w, line 256: This is the fallback for general `RuntimeException`, so we don't return a 500. I also used the `error` log level because this handler should never be invoked.

docs/Signer-Cloud-API.md
```diff
@@ -18,10 +18,23 @@ Signer Cloud Server provides a RESTful API that allows to control specific parts
 
 Signer Cloud Server uses following format for error response body, accompanied by an appropriate HTTP status code. Besides the HTTP error codes that application server may return regardless of server application (such as 404 when resource is not found or 503 when server is down), following ERROR codes may be returned:
 
-| Error Code         | HTTP Code | Description                                                |
-|:-------------------|:----------|:-----------------------------------------------------------|
-| ERROR_GENERIC      | 400       | Issue with a request format or issue of the business logic |
-| ERROR_UNAUTHORIZED | 401       | Unauthorized request                                       |
+| Error Code                       | HTTP Code | Description                                                                                                |
+|:---------------------------------|:----------|:-----------------------------------------------------------------------------------------------------------|
+| ERROR_UNAUTHORIZED               | 401       | Unauthorized request                                                                                       |
```
michal-rozehnal-w: Added more error codes so the user can identify more precisely what went wrong. I will confirm this with @tech-kratas

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentController.java
```diff
@@ -65,7 +66,10 @@ UploadDocumentResponse upload(
             @RequestParam("file") final MultipartFile file
     ) throws Throwable {
         logger.info("action: uploadDocument, state: initiated, externalSignerId: {}, externalDocumentId: {}", externalSignerId, externalDocumentId);
```
michal-rozehnal-w, line 68: This same pattern is used in most of the methods in the controllers:
```
logger.info("start");
final var result = Try.execute(...);
if (result.isSuccess()) {
  logger.info("success");
  return result;
} else {
  logger.info("fail");
  return error;
}
```
I considered extracting it into a separate method with lambdas and log messages as parameters, but I wasn’t sure about the benefit. In particular, since we usually use different parameters for log messages, having them as lambdas could reduce readability. So I left the code as it is.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/Try.java
```diff
@@ -17,14 +17,16 @@
...
+
 /**
  * Wrapper for the result of an operation from the service layer.
  *
  * @author Michal Rozehnal, michal.rozehnal@wultra.com
  */
 public sealed interface Try<T> permits Try.TryError, Try.TrySuccess{
     boolean isSuccess();
-    Throwable getError();
+    RuntimeException getError();
```
michal-rozehnal-w, line 29: Using `Throwable` was too broad here. I also considered `Error`, but since all our custom exceptions are `RuntimeException`s and we don't throw checked exceptions, I used that type instead.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/CertificateRevocationService.java
```diff
@@ -67,15 +67,15 @@ void revokeCertificate(final IssuedCertificateMetadata certificateMetadata, fina
             setRevokedStatus(certificateMetadata);
             logger.info("Certificate successfully revoked. Issuer DN: {}, Serial Number: {}", issuerDn, serialNumber);
         } catch (final RestClientException e) {
-            logger.warn("Exception when revoking certificate: {}", e.getResponse(), e);
+            logger.warn("Error from EJBCA server when revoking certificate: {}", e.getResponse(), e);
```
michal-rozehnal-w: For the `RestClientException` it is good to log the response, because it provides more details. The `.getMessage()` returns just HTTP status code.

docs/Signer-Cloud-API.md
```diff
@@ -75,19 +88,10 @@ Create new signer and enroll for new certificate using CSR. System will track ce
 
 #### Response 200
 
-```json
-{
-  "result": "String",
-  "resultReason": null
-}
+```
+200 OK
```
michal-rozehnal-w, line 90: To keep all endpoints consistent, it now returns 400 with the proper error code and description in case of failure.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerServiceTest.java
```diff
@@ -119,7 +119,7 @@ class SignerServiceTest {
     private ArgumentCaptor<IssuedCertificateMetadata> issuedCertificateCaptor;
 
     @BeforeEach
-    void setUp() throws CertificateException, IOException {
+    void setUp() throws CertificateException {
```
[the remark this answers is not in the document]
michal-rozehnal-w: I think it is not true. The `CertificateUtils.base64ToX509Certificate` throws only `CertificateException` and since `IOException` is checked, it would require to be part of the method signature, and it is not.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -17,6 +17,8 @@
  */
 package com.wultra.signercloud.server.signer;
 
+import com.wultra.core.rest.client.base.RestClientException;
```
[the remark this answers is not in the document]
michal-rozehnal-w: It is used. There is 9 usages in the class.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerNotFoundException.java
```diff
@@ -23,7 +23,11 @@
...
  */
 public class SignerNotFoundException extends RuntimeException {
-    public SignerNotFoundException(String message) {
-        super(message);
+    public SignerNotFoundException(final String externalSignerId) {
+        super("Signer with ID %s not found".formatted(externalSignerId));
+    }
+
+    public SignerNotFoundException(final Long signerId) {
+        super("Signer with internal ID %d not found".formatted(signerId));
```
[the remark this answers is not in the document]
michal-rozehnal-w: Well spotted. Fixed by using the `%s`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -108,7 +104,7 @@ CleanupResult cleanupDocuments() {
 
     private void processRetention(final DocumentStatus status, final Duration retentionPeriod, final Consumer<String> resultConsumer, final Instant now) {
         if (retentionPeriod != null) {
-            long deletedCount = documentRepository.deleteByStatusAndTimestampCreatedBefore(status, now.minus(retentionPeriod));
+            final long deletedCount = documentRepository.deleteByStatusAndTimestampCreatedBefore(status, now.minus(retentionPeriod));
```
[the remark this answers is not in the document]
michal-rozehnal-w: I saw a warning in IntelliJ, so I fixed it. It's such a tiny change that I think it can be included in this PR.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/callback/CallbackService.java
```diff
@@ -271,7 +271,7 @@ private void postCallback(final CallbackEventData callbackEventData) {
                     onError);
 
             logger.debug("CallbackEvent {} was dispatched.", callbackEventData.id());
-        } catch (RestClientException e) {
+        } catch (final RestClientException e) {
```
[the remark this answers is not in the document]
michal-rozehnal-w: Same as above. It's such a tiny change that I think it can be included in this PR.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -47,26 +49,27 @@ class SignerController {
...
             logger.info("action: createUpdateSigner, state: succeeded");
-            return new SignerResponse(SignerResponseResult.OK, null);
         } else {
-            logger.error("action: createUpdateSigner, state: failed, errorMessage: {}", result.getError().getMessage());
-            return new SignerResponse(SignerResponseResult.FAIL, result.getError().getMessage());
+            final var error = result.getError();
+            logger.error("action: createUpdateSigner, state: failed, errorMessage: {}", error.getMessage());
+            throw error;
         }
     }
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is not exactly true, because `Try::execute` doesn’t throw an exception; it returns an error result. If the service is called directly and an exception is thrown, the else block will never be reached. So usage of `Try` is on purpose and have benefits.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -77,26 +80,27 @@ SignerResponse createUpdate(@Valid @RequestBody final CreateUpdateSignerRequest
...
             logger.info("action: updateSignerStatus, state: succeeded");
-            return new SignerResponse(SignerResponseResult.OK, null);
         } else {
-            logger.error("action: updateSignerStatus, state: failed, errorMessage: {}", result.getError().getMessage());
-            return new SignerResponse(SignerResponseResult.FAIL, result.getError().getMessage());
+            final var error = result.getError();
+            logger.error("action: updateSignerStatus, state: failed, errorMessage: {}", error.getMessage());
+            throw error;
         }
     }
```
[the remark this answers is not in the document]
michal-rozehnal-w: Same as above

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerController.java
```diff
@@ -116,16 +120,19 @@ SignerResponse updateStatus(@PathVariable final String externalSignerId, @Valid
...
             logger.info("action: getSignerDetail, state: succeeded");
             return result.getResponse();
         } else {
-            logger.info("action: getSignerDetail, state: failed, errorMessage: {}", result.getError().getMessage());
-            throw result.getError();
+            final var error = result.getError();
+            logger.info("action: getSignerDetail, state: failed, errorMessage: {}", error.getMessage());
+            throw error;
         }
     }
```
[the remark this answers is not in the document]
michal-rozehnal-w: Same as above

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerServiceTest.java
```diff
@@ -420,94 +414,88 @@ void testCreateUpdateSignerWhenSignerIsUpdatedThenIssuedCertificatedIsSaved() th
...
 
         when(signerRepository.findByExternalSignerId(EXTERNAL_SIGNER_ID)).thenReturn(Optional.of(signer));
 
         // when
-        final var response = signerService.updateStatus(
-                EXTERNAL_SIGNER_ID,
-                new UpdateSignerStatusRequest(SignerStatus.ACTIVE, null)
+        final var exception = assertThrows(
+                SignerStatusTransitionException.class,
+                () -> signerService.updateStatus(EXTERNAL_SIGNER_ID, new UpdateSignerStatusRequest(SignerStatus.BLOCKED, null))
```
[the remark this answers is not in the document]
michal-rozehnal-w: I think the test is correct. Its name is `testUpdateStatusWhenStatusIsSetToRevokedThenEjbcaIsCalled` and the request is `new UpdateSignerStatusRequest(SignerStatus.REVOKED, null)`

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentContentNotFoundException.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package com.wultra.signercloud.server.document;
+
+/**
+ * Exception thrown when a {@link DocumentContent} is not found in the system.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+public class DocumentContentNotFoundException extends RuntimeException {
+    public DocumentContentNotFoundException(final String documentUuid) {
```
banterCZ: 🤔 Since `documentUuid` is a type of String, it is confusing, because you may expect, that it is the whole message.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentNotFoundException.java
```diff
@@ -18,12 +18,12 @@
...
- * Exception thrown when a {@link Document} or {@link DocumentContent} is not found in the system.
+ * Exception thrown when a {@link Document} is not found in the system.
  *
  * @author Michal Rozehnal, michal.rozehnal@wultra.com
  */
 public class DocumentNotFoundException extends RuntimeException {
-    public DocumentNotFoundException(final String message) {
-        super(message);
+    public DocumentNotFoundException(final String documentUuid) {
+        super("Document with ID %s not found".formatted(documentUuid));
```
banterCZ: ditto

signer-cloud-server/src/test/java/com/wultra/signercloud/server/signer/SignerServiceTest.java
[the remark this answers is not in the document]
banterCZ: I don't mind, for tests, I use just `Exception` for all.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerNotFoundException.java
```diff
@@ -23,7 +23,11 @@
  * @author Michal Rozehnal, michal.rozehnal@wultra.com
  */
 public class SignerNotFoundException extends RuntimeException {
-    public SignerNotFoundException(String message) {
-        super(message);
+    public SignerNotFoundException(final String externalSignerId) {
```
banterCZ: ditto

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -201,18 +183,19 @@ private static byte[] getFileBytes(final MultipartFile file) {
         try {
             return file.getBytes();
         } catch (final IOException e) {
-            throw new DocumentUploadException("Failed to read file: " + e.getMessage());
+            logger.warn("Exception when reading upload file", e);
```
banterCZ: 🤔 I would expect that the exception is logged in the exception handler. Isn't it logged twice, now? IMHO it is a good approach to have the handler as the last resort and be sure that the exception is logged. Have you considered that?

banterCZ (commented):
A good job.
I consider confusing passing String IDs to exception constructors.
Let's think about where to log the exception.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentContentNotFoundException.java
michal-rozehnal-w: Good point — you are absolutely right that most exceptions expect the full message in the constructor with a String parameter. I created static methods for IDs and made the constructor private.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentNotFoundException.java
michal-rozehnal-w: Fixed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/signer/SignerNotFoundException.java
michal-rozehnal-w: Fixed

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
michal-rozehnal-w: All exception logging for the REST API is now handled in the handler. This is probably a cleaner approach because it helps fulfill our logging rule: _“If it’s worth saving to a log file, it’s worth showing in the user interface.”_ 
We just need to be mindful that the handler only works for REST API calls; for other logic (such as jobs), we still need to use the logger.

docs/Signer-Cloud-API.md
```diff
@@ -18,10 +18,21 @@ Signer Cloud Server provides a RESTful API that allows to control specific parts
...
+| REQUEST_VALIDATION_ERROR         | 400       | REST API endpoint called with invalid body or parameters                                                                                  |
+| ERROR_RESOURCE_NOT_FOUND         | 400       | Resource is not found                                                                                                                     |
+| CERTIFICATE_PROCESSING_ERROR     | 500       | Issue with processing the certificate                                                                                                     |
+| CSR_INVALID_SIGNATURE_ERROR      | 400       | Error when signature of CSR (Certificate Signing Request) is invalid                                                                      |
+| CSR_SIGNATURE_VERIFICATION_ERROR | 500       | Error when signature of CSR (Certificate Signing Request) could not be verified. Can indicates problem with Power Auth server             |
+| CERTIFICATE_AUTHORITY_ERROR      | 400       | Error returned from Certificate Authority. Can indicate problem with Certificate Authority server                                         |
```
michal-rozehnal-w: I discussed these changes with @tech-kratas . The tricky part is `CERTIFICATE_AUTHORITY_ERROR`, which represents any error with EJBCA. It can return both 500s and 400s. I think we should avoid mirroring EJBCA status codes—for example, it returns 404 when a resource is not found, but we use 400. So I used 400 here but I am open to any suggestions.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
banterCZ: Well, mind that sometimes it is worth saving to a log file, but from the security point of view, you do not want to reveal details at the user interface.

docs/Signer-Cloud-API.md
banterCZ: From my point of view, 500 is too generous. What about using 503 instead?

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/RestExceptionHandler.java
```diff
@@ -74,59 +113,149 @@ public ResponseEntity<ErrorResponse> handleSignerNotFoundException(final SignerN
...
+    /**
+     * Handler for {@link SignerStatusTransitionException} producing {@link HttpStatus#BAD_REQUEST} response.
+     *
+     * @param ex the exception
+     * @return response as {@link ResponseEntity}
+     */
+    @ExceptionHandler
+    public ResponseEntity<ErrorResponse> handleSignerStatusTransitionException(final SignerStatusTransitionException ex) {
+        logger.warn("Illegal Signer status transition", ex);
+        return buildErrorResponse(HttpStatus.BAD_REQUEST, ErrorCode.ILLEGAL_OPERATION_ERROR, ex.getMessage());
```
banterCZ, line 245: From the security point of view, you do not always want to reveal details at the user interface. But in the Context of Signer Cloud, I do not see any issue yet.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentContentNotFoundException.java
```diff
@@ -0,0 +1,40 @@
...
+        super(message);
+    }
+
+    /**
+     * Creates exception for document content.
+     *
+     * @param documentUuid Document UUID
+     * @return Exception instance
+     */
+    public static DocumentContentNotFoundException forId(final String documentUuid) {
```
banterCZ, line 36: 👏 A nice idea!

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/RestExceptionHandler.java
banterCZ: 🤔 Masquerading internal error to 400 instead of 500 could be misleading. 400 should point out that something is wrong with the client's request.

banterCZ (approved):

docs/Signer-Cloud-API.md
michal-rozehnal-w: Already discussed and mentioned in another comment

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/RestExceptionHandler.java
michal-rozehnal-w: Thanks for pointing that out. You are absolutely right — it is sometimes tricky to find the balance between providing enough details to the user while not revealing too much for security reasons. I also agree that, at this point, the Signer Cloud works mostly with public data (CSR, certificate, etc.); however, this could change in the future, and we must always keep it in mind

docs/Signer-Cloud-API.md
michal-rozehnal-w: I adjusted the mapping.
`CSR_INVALID_SIGNATURE_ERROR`: always 503
`CERTIFICATE_AUTHORITY_ERROR`: 5xx -> 503, 4xx -> 400

signer-cloud-server/src/main/java/com/wultra/signercloud/server/restapi/RestExceptionHandler.java
michal-rozehnal-w: Yep, changed to `500`.

PR #102, opened 2025-10-01
Fix pades signature context consistency

michal-rozehnal-w:


signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -276,27 +279,39 @@ private byte[] verifySignatureAndSignDocument(
...
             final String hashSignatureBase64,
             final byte[] documentBytes,
-            final DigestAlgorithm signatureAlgorithm) {
+            final Document document) {
 
         final var certificate = convertCertificate(signer);
         final var certificateChain = convertCertificateChain(signer);
 
-        final var signatureParams = createSignatureParameters(certificate, signatureAlgorithm, certificateChain);
+        final var signatureParams = createSignatureParameters(certificate, certificateChain);
```
michal-rozehnal-w, line 287: This is important part. We must use exactly same params as for computing hash. We don't have to set the `deterministicId` because is it calculated from the `setSigningDate`. So for same Date it computes same `deterministicId`

michal-rozehnal-w:
This was just quick test of the fix. Here is PR where the issue was fixed https://github.com/wultra/signer-cloud/pull/104

PR #104, opened 2025-10-02
Fix #103: Invalid signature in signed pdf

### LINKED ISSUE #103: Invalid signature in signed PDF
### Affected product(s) and version(s)

_No response_

### Environment

_No response_

### Steps to reproduce, current result, expected result

Sign and download PDF

### Log output

When signed PDF is opened in Adobe Acrobat Reader there is a message that document was changed.

Image

The signature must be valid.

michal-rozehnal-w:


signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -19,14 +19,14 @@
 
 import com.wultra.signercloud.server.signer.*;
 import eu.europa.esig.dss.enumerations.DigestAlgorithm;
+import eu.europa.esig.dss.model.DSSDocument;
```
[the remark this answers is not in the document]
michal-rozehnal-w: Since this is test it should be ok. Otherwise it won't pass.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -397,7 +401,7 @@ void testSignDocumentWhenSignatureIsNotValidThenExceptionIsThrown() {
         when(documentContentRepository.findById(AggregateReference.to(DOCUMENT_CONTENT_ID))).thenReturn(Optional.of(documentContent));
         when(signerRepository.findById(AggregateReference.to(SIGNER_ID))).thenReturn(Optional.of(signer));
         when(documentConfigurationProperties.getWaiting()).thenReturn(waitingDuration);
-        when(documentConfigurationProperties.getSignatureHashAlgorithm()).thenReturn(DigestAlgorithm.SHA384);
+        when(documentConfigurationProperties.getContentHashAlgorithm()).thenReturn(DigestAlgorithm.SHA256);
```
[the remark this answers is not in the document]
michal-rozehnal-w: I think it’s better to duplicate this in the tests than to put it in `@BeforeEach` or another common method. Not all tests use this mocking, and using `lenient` isn’t ideal either.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -434,7 +438,9 @@ void testSignDocumentWhenSignatureIsValidThenSuccessResultIsReturned() {
...
```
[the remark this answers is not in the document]
michal-rozehnal-w: As above

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -475,7 +481,9 @@ void testSignDocumentWhenCertificateChainIsSetThenPadesServiceIsCalledWithTheCha
...
```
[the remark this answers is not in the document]
michal-rozehnal-w: As above

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -41,14 +39,10 @@
...
 import java.io.IOException;
-import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is default IntelliJ formatting. We don't have any our rule about using `*` for imports.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -187,15 +183,16 @@ private static byte[] getFileBytes(final MultipartFile file) {
...
+    private String computeHash(final byte[] content, final Signer signer, final Instant timestampSigned) {
+        final var certificate = convertCertificate(signer);
+        final var certificateChain = convertCertificateChain(signer);
+
+        final var signatureParams = createSignatureParameters(certificate, certificateChain, timestampSigned);
+
+        final var document = new InMemoryDocument(content);
+        final var toSignBytes = padesService.getDataToSign(document, signatureParams);
+
+        return Base64.getEncoder().encodeToString(toSignBytes.getBytes());
```
michal-rozehnal-w, line 195: When I implemented this the first time, I misunderstood it. For PAdES, we must compute the hash from both the document and the signature, so the previous implementation was not correct.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -324,17 +320,39 @@ private static List<CertificateToken> convertCertificateChain(final Signer signe
...
     ) {
         final var params = new PAdESSignatureParameters();
-        params.setDigestAlgorithm(algorithm);
+        params.setDigestAlgorithm(configurationProperties.getHashAlgorithm());
         params.setSignatureLevel(SignatureLevel.PAdES_BASELINE_B);
         params.setSigningCertificate(certificateToken);
         params.setCertificateChain(certificateChain);
 
+        params.bLevel().setSigningDate(Date.from(timestampSigned));
+        params.setSigningTimeZone(TimeZone.getTimeZone("UTC"));
```
michal-rozehnal-w, line 354: This took me a lot of debugging because the integration tests passed locally but failed on GitHub. The problem was that if `setSigningTimeZone` is not set, the server’s default time zone is used. Since my local machine is not in UTC but GitHub is, the same Instant produced a different hash.

signer-cloud-server/src/main/resources/application.properties
```diff
@@ -39,8 +39,7 @@ signer-cloud.server.document.waiting.timeout=${SIGNER_CLOUD_DOCUMENT_WAITING_TIM
 signer-cloud.server.document.rejected.retention-period=${SIGNER_CLOUD_DOCUMENT_REJECTED_RETENTION_PERIOD:}
 signer-cloud.server.document.signed.retention-period=${SIGNER_CLOUD_DOCUMENT_SIGNED_RETENTION_PERIOD:}
 
-signer-cloud.server.document.content-hash-algorithm=SHA256
-signer-cloud.server.document.signature-hash-algorithm=SHA384
+signer-cloud.server.document.hash-algorithm=SHA256
```
michal-rozehnal-w: We only need to specify the algorithm for computing the hash. The signature algorithm is defined in the certificate.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentControllerIntTest.java
```diff
@@ -290,29 +288,6 @@ void testSignWhenDocumentIsNotInWaitingStateThen400WithCorrectResponseIsReturned
         assertErrorResponse(responseBody, ErrorCode.ILLEGAL_OPERATION_ERROR, "Document is not in state when it can be signed");
     }
 
-    @Test
-    void testSignWhenDocumentSignAttemptIsAfterDeadlineThen400WithCorrectResponseIsReturned() throws Exception {
```
michal-rozehnal-w, line 294: I removed this test because we now use a fixed value in `DOCUMENT_TIMESTAMP_CREATED` (instead of `Instant.now()`) to generate a deterministic hash. With this approach, `padesService` is used for real verification that the hash and signature match in these integration tests. However, it is not possible to mock the current Instant. We could create a wrapper `@Service` for obtaining the current `Instant` to allow mocking. Nevertheless, we still have a unit test for this scenario, so having an additional integration test is probably unnecessary.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -62,18 +62,19 @@ class DocumentServiceTest {
 
     private static final long SIGNER_ID = 1L;
     private static final String EXTERNAL_SIGNER_ID = "756419e1-1d85-4172-815d-d8653ecd3a89";
-    private static final String CERTIFICATE = "MIICFDCCAZugAwIBAgIUC0O75BZKicH8bDlRUBgC8h7bTdgwCgYIKoZIzj0EAwMwFDESMBAGA1UEAwwJSXNzdWluZ0NBMB4XDTI1MDgyNzA3NTUyOVoXDTI3MDgxMTA5MTQ0NlowNjERMA8GA1UEAwwISm9obiBEb2UxFDASBgNVBAoMC0V4YW1wbGVDb3JwMQswCQYDVQQGEwJVUzB2MBAGByqGSM49AgEGBSuBBAAiA2IABENmfWbIOOXtmGfPQdN7oY8+y0pF7rliVvg709N7X3Q/uW5g7rTO7JWc4BPX8+rGjbkaAU7tO6nt0JCaBtYlwzmGLkwywafNd4iTsFaZRbFLNdZue/c0poVN9RmwLTOH/6OBizCBiDAMBgNVHRMBAf8EAjAAMB8GA1UdIwQYMBaAFJ0dk1DJP8vLqD/Dx15EMOEpmqkOMCgGA1UdJQQhMB8GCCsGAQUFBwMCBggrBgEFBQcDBAYJKoZIhvcvAQEFMB0GA1UdDgQWBBT9ooCMF8fza6pXM3q4FQrfJ0nkhjAOBgNVHQ8BAf8EBAMCBeAwCgYIKoZIzj0EAwMDZwAwZAIwaeS/siF1g5vbaNXrnQM9xJOQmUG92HyNOCTKh/x1PA9b/VwtpodSjkIOiOxJQ56aAjBQit9XczUVNp5qGdrLO3Ac730VokRvphNBtupJbdnkpywejktZi00LM8MsuZA7Piw=";
+    // private key: "Pw7qJc9ZFEMakSuAZEBYPq+0j0iuXkvqqWQzbyaNiGw="
```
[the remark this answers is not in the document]
michal-rozehnal-w: This is a randomly generated certificate, purely for testing purposes. I included the private key in case we need to adjust any values later.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentControllerIntTest.java
```diff
@@ -87,23 +88,23 @@ class DocumentControllerIntTest {
...
-    private static final String CERTIFICATE = "MIICFDCCAZugAwIBAgIUC0O75BZKicH8bDlRUBgC8h7bTdgwCgYIKoZIzj0EAwMwFDESMBAGA1UEAwwJSXNzdWluZ0NBMB4XDTI1MDgyNzA3NTUyOVoXDTI3MDgxMTA5MTQ0NlowNjERMA8GA1UEAwwISm9obiBEb2UxFDASBgNVBAoMC0V4YW1wbGVDb3JwMQswCQYDVQQGEwJVUzB2MBAGByqGSM49AgEGBSuBBAAiA2IABENmfWbIOOXtmGfPQdN7oY8+y0pF7rliVvg709N7X3Q/uW5g7rTO7JWc4BPX8+rGjbkaAU7tO6nt0JCaBtYlwzmGLkwywafNd4iTsFaZRbFLNdZue/c0poVN9RmwLTOH/6OBizCBiDAMBgNVHRMBAf8EAjAAMB8GA1UdIwQYMBaAFJ0dk1DJP8vLqD/Dx15EMOEpmqkOMCgGA1UdJQQhMB8GCCsGAQUFBwMCBggrBgEFBQcDBAYJKoZIhvcvAQEFMB0GA1UdDgQWBBT9ooCMF8fza6pXM3q4FQrfJ0nkhjAOBgNVHQ8BAf8EBAMCBeAwCgYIKoZIzj0EAwMDZwAwZAIwaeS/siF1g5vbaNXrnQM9xJOQmUG92HyNOCTKh/x1PA9b/VwtpodSjkIOiOxJQ56aAjBQit9XczUVNp5qGdrLO3Ac730VokRvphNBtupJbdnkpywejktZi00LM8MsuZA7Piw=";
+    // private key: "AJX0rDTQYUR2oXjJPkMAfECh8P2mb2S2PweW3Lo/UhIw"
```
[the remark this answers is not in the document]
michal-rozehnal-w: Same as above

signer-cloud-server/src/main/resources/application.properties
banterCZ: Shouldn't we expose that as ENV too? cc @korbelm

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -62,18 +62,19 @@ class DocumentServiceTest {
...
...wMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMD4KL0J5dGVSYW5nZSBbMCA0NzggMTk0MjQgNzI5XSAgICAgICAgICAgICAgICAgICAKPj4KZW5kb2JqCjMgMCBvYmoKPDwKL1R5cGUgL1BhZ2UKL1BhcmVudCAyIDAgUgovTWVkaWFCb3ggWzAgMCAxMCAxMF0KL0Fubm90cyBbNCAwIFJdCj4+CmVuZG9iago2IDAgb2JqCjw8Ci9OIDcgMCBSCj4+CmVuZG9iago3IDAgb2JqCjw8Ci9MZW5ndGggMAovVHlwZSAvWE9iamVjdAovU3VidHlwZSAvRm9ybQovQkJveCBbMC4wIDAuMCAwLjAgMC4wXQo+PgpzdHJlYW0NCg0KZW5kc3RyZWFtCmVuZG9iagoxIDAgb2JqCjw8Ci9UeXBlIC9DYXRhbG9nCi9QYWdlcyAyIDAgUgovQWNyb0Zvcm0gPDwKL0ZpZWxkcyBbNCAwIFJdCi9TaWdGbGFncyAzCj4+Cj4+CmVuZG9iagp4cmVmCjAgMgowMDAwMDAwMDAwIDY1NTM1IGYNCjAwMDAwMTk3MDMgMDAwMDAgbg0KMyA1CjAwMDAwMTk0ODMgMDAwMDAgbg0KMDAwMDAwMDIyNyAwMDAwMCBuDQowMDAwMDAwMzYyIDAwMDAwIG4NCjAwMDAwMTk1NjggMDAwMDAgbg0KMDAwMDAxOTU5OCAwMDAwMCBuDQp0cmFpbGVyCjw8Ci9Sb290IDEgMCBSCi9JRCBbPDhBQkU4MjdCRjBEQkQzQjQ2RjJBQzhEQjEwRjFCNUUzQ0EwNjYyMEI3RUQ3MzJEQ0Q4NzU0MDBBRjUwNTA5NDA+IDw4QUJFODI3QkYwREJEM0I0NkYyQUM4REIxMEYxQjVFM0NBMDY2MjBCN0VENzMyRENEODc1NDAwQUY1MDUwOTQwPl0KL1ByZXYgMAovU2l6ZSA4Cj4+CnN0YXJ0eHJlZgoxOTc5NgolJUVPRgo=");
```
banterCZ: 🤔 Maybe this could be externalized to a file.

banterCZ (approved):
A nice result of debugging!

signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
michal-rozehnal-w: Good idea. In IntelliJ it’s just one line, but here I can see it’s quite large. I used the already existing files from the test resources.

signer-cloud-server/src/main/resources/application.properties
michal-rozehnal-w: Valid point. I think it depends on whether we want to make this value more visible for configuration. For example, we intentionally have `signer-cloud.server.security.auth.type=OAUTH2`, since we don’t want to offer `BASIC_HTTP` except for development. Also we haven’t tested any value other than SHA256, but it should work with any value from the `DigestAlgorithm` enum (at least I don't see any obvious issue with different value here). If we really want a variable for it, please let me know and I will add it.

michal-rozehnal-w: I added `SIGNER_CLOUD_DOCUMENT_HASH_ALGORITHM`, default value is `SHA256`

banterCZ: We may open an extra issue for SHA384

PR #113, opened 2025-10-08
Fix #107: Add support for pades B-T signature level

### LINKED ISSUE #107: Add Support for PAdES-B-T
### Description

Add support for PAdES-B-T signature (with signature timestamp).

Step 1.
Start with POC using [FreeTSA](https://www.freetsa.org/) Time Stamp Authority.

Step 2.
Integrate it into product. TSA should be configurable (it can be simple config but we may need  to implement some kind "connector" in case there are differences in API for various TSA's).

We need to add possibility to choose which signature type we want to use. There should be "default signature type" in server config as well as possibility to overwrite default value in API request (TBC).

### Acceptance criteria

Signature is valid PAdES-B-T according to [DSS Validator](https://ec.europa.eu/digital-building-blocks/DSS/webapp-demo/validation).

### Technical specification

[ETSI EN 319 142-1](https://www.etsi.org/deliver/etsi_en/319100_319199/31914201/01.02.01_60/en_31914201v010201p.pdf) - Chapter 6

[DSS - Creating a baseline T signature](https://ec.europa.eu/digital-building-blocks/DSS/webapp-demo/doc/dss-documentation.html#_creating_a_baseline_t_signature)
 
[DSS - Augmenting an AdES baseline B signature](https://ec.europa.eu/digital-building-blocks/DSS/webapp-demo/doc/dss-documentation.html#_augmenting_an_ades_baseline_b_signature)

### QA specification

- Generate signed document
- Use DSS Validator
- Check results

michal-rozehnal-w:


signer-cloud-server/src/test/java/com/wultra/signercloud/server/document/DocumentServiceTest.java
```diff
@@ -514,6 +557,181 @@ void testSignDocumentWhenCertificateChainIsSetThenPadesServiceIsCalledWithTheCha
...
+        when(pAdESServiceConfig.getTsaUrl()).thenReturn("https://freetsa.org/tsr");
+        when(pAdESServiceConfig.getHashAlgorithm()).thenReturn(DigestAlgorithm.SHA256);
+        when(pAdESService.getDataToSign(any(DSSDocument.class), any(PAdESSignatureParameters.class)))
+                .thenReturn(new ToBeSigned(Base64.getDecoder().decode(DOCUMENT_HASH)));
+        when(pAdESService.isValidSignatureValue(any(ToBeSigned.class), any(SignatureValue.class), any(CertificateToken.class)))
+                .thenReturn(true);
+        when(pAdESService.signDocument(any(InMemoryDocument.class), any(PAdESSignatureParameters.class), any(SignatureValue.class)))
+                .thenReturn(new InMemoryDocument(signedDocumentContent));
+
+
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

docs-private/How-to-test-document-signing.md
```diff
@@ -884,4 +888,19 @@ curl -X POST "https://login.microsoftonline.com/09456bd9-6d0f-40e3-b972-fb499273
...
 ```
 
-Values for `<CLIENT_ID>` and `<CLIENT_SECRET>` are stored in password manager.
\ No newline at end of file
+Values for `<CLIENT_ID>` and `<CLIENT_SECRET>` are stored in password manager.
+
+
+### Document signature level
+
+By default, the **PAdES B-B** level is used for signatures. Another supported level is **PAdES B-L**.
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

docs-private/signer_cloud_server_database_schema.puml
```diff
@@ -30,6 +30,7 @@ class sc_document <<table>> {
     hash varchar(255) NN
     status varchar(50) NN I (WAITING/REJECTED/SIGNED)
     signature varchar(255)
+    signature_level varchar(50)
```
michal-rozehnal-w, line 33: This change was requested by @tech-kratas to make it immediately clear what the signature level is. This information is also stored in signed document.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/PAdESServiceConfig.java
```diff
@@ -27,12 +34,40 @@
...
+     * Default signature for the document.
+     */
+    @Value("${signer-cloud.server.pades.signature-level}")
+    private DocumentSignatureLevel signatureLevel;
+
+    /**
+     * Algorithm used to compute the hash of the document for signing.
+     */
+    @Value("${signer-cloud.server.pades.hash-algorithm}")
+    private DigestAlgorithm hashAlgorithm;
```
michal-rozehnal-w: I moved this one from `DocumentConfigurationProperties`. IMHO it makes more sense to have it here when we have other pades config values.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/document/DocumentService.java
```diff
@@ -187,7 +190,12 @@ private String computeHash(final byte[] content, final Signer signer, final Inst
         final var certificate = convertCertificate(signer);
         final var certificateChain = convertCertificateChain(signer);
 
-        final var signatureParams = createSignatureParameters(certificate, certificateChain, timestampSigned);
+        final var signatureParams = createSignatureParameters(
+                certificate,
+                certificateChain,
+                timestampSigned,
+                pAdESServiceConfig.getSignatureLevel()
```
michal-rozehnal-w: It doesn't matter what level is set here. I tested `PADES_B_T` and then sign the document with the level `PADES_B_B`, and it works fine. At least this is true for these two levels, since the timestamp is an extra unsigned layer. I am not sure about the more advanced levels (B-LT and B-LTA), but we don't support them at the moment.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/PAdESServiceConfig.java
banterCZ: 🤔 I would rather keep a separate concept of `*Config` classes producing beans and `*ConfigurationProperties` classes with a prefix (not using `@Value`), keeping the configuration itself. Moving in general is fine.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/PAdESServiceConfig.java
```diff
@@ -27,12 +34,40 @@
...
 
     @Bean
     public PAdESService padesService() {
-        return new PAdESService(new CommonCertificateVerifier());
+        final var padesService = new PAdESService(new CommonCertificateVerifier());
+
+        if (StringUtils.isNotEmpty(tsaUrl)) {
+            final var tspSource = new OnlineTSPSource(tsaUrl);
```
banterCZ: Don't we want to log it? 
It does not have to be implemented, just an idea. What about making `tspSource` as a conditional bean and pasting it as `padesService(@Autowired(required = false) OnlineTSPSource tspSource)`?

banterCZ (commented):
Good job. But let's discuss `ConfigurationProperties` versus `Config` classes first.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/PAdESServiceConfig.java
banterCZ: Thanks for the log. I suggest moving it, just to be safe, if the constructor fails.
```java
logger.info("Setting TSA URL: {}", tsaUrl);
final var tspSource = new OnlineTSPSource(tsaUrl);
```

banterCZ (approved):
Thanks for extracting the ConfigurationProperties.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/PAdESServiceConfig.java
michal-rozehnal-w: Fixed — I also added the conditional bean. I wanted to push both changes at once, but I accidentally pushed only the first one.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/PAdESServiceConfig.java
michal-rozehnal-w: Good idea, I moved the properties into separate class

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/OnlineTSPSourceConfig.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Configuration for TSA source.
+ *
+ * @author Michal Rozehnal, michal.rozehnal@wultra.com
+ */
+@Configuration
+@AllArgsConstructor
+@Slf4j
+public class OnlineTSPSourceConfig {
```
banterCZ: Nitpick: Does it have to be _public_?

michal-rozehnal-w: Right, it is not necessary. I removed it

PR #135, opened 2025-10-22
Fix #132: Use dynamically generated resources in integration tests

### LINKED ISSUE #132: TechDebt: Use dynamically generated certificates in tests
### Task specification

Currently, we use constants for certificates in integration tests. This is fine for now, but when the certificate expires, a few tests will fail because the pades service has some internal validations. We can generate a certificate for each test suite and temporarily store it so we can retrieve it for debugging if any test fails.

michal-rozehnal-w:


.github/workflows/maven-test.yml
```diff
@@ -14,7 +14,8 @@ on:
 
 jobs:
   maven-tests:
-    uses: wultra/wultra-infrastructure/.github/workflows/maven-test.yml@develop
+    uses: wultra/wultra-infrastructure/.github/workflows/maven-test.yml@issues/persist-maven-test-resource
```
michal-rozehnal-w: - [x] Change branch to `develop` when the [PR](https://github.com/wultra/wultra-infrastructure/pull/62) in infrastructure repo is merged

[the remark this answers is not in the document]
michal-rozehnal-w: I have task for it, will be changed to `develop` when the corresponding PR is merged.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
```diff
@@ -0,0 +1,315 @@
...
+        final var userCsrPem = stringWriter.toString();
+
+        final var userCriDer = userCsr.toASN1Structure()
+                .getCertificationRequestInfo()
+                .getEncoded();
+
+        final var userCertificate = (X509Certificate) keyStore.getCertificate("user");
+        final var userCertificateChain = Arrays.stream(keyStore.getCertificateChain("user"))
+                .filter(c -> !c.equals(userCertificate))
+                .map(c -> (X509Certificate) c )
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
```diff
@@ -0,0 +1,315 @@
...
+    public static IntTestResources prepare() throws Exception {
+        if (testResources == null) {
+            createResources();
+        }
+
+        return testResources;
+    }
+
+    private static void createResources() throws Exception {
+        Security.addProvider(new BouncyCastleProvider());
```
banterCZ: 🤔 This is already in `SignerCloudServerApplication` but not in the static constructor. Maybe we may move that.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
```diff
@@ -0,0 +1,315 @@
...
+    private static IntTestResources testResources;
+
+    /**
+     * Generates and stores resources (keystore, CSR, signed documents) needed for integration tests.
+     *
+     * @return generated resources
+     * @throws Exception when generation or storing of resources fails
+     */
+    public static IntTestResources prepare() throws Exception {
+        if (testResources == null) {
```
banterCZ, line 73: 🤔 Is it threat safe? But we do not run the test in parallel yet, right?

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
```diff
@@ -0,0 +1,315 @@
...
+                .getEncoded();
+
+        final var userCertificate = (X509Certificate) keyStore.getCertificate("user");
+        final var userCertificateChain = Arrays.stream(keyStore.getCertificateChain("user"))
+                .filter(c -> !c.equals(userCertificate))
+                .map(c -> (X509Certificate) c)
+                .toList();
+
+
+        return new SignerResources(
```
banterCZ: 🤔 Ok, but isn't it safer to use a builder?

.github/workflows/maven-test.yml
```diff
@@ -14,7 +14,8 @@ on:
...
     secrets: inherit
     with:
-      java_version: 21
\ No newline at end of file
+      java_version: 21
+      artifacts_path: '**/signer-cloud-server/target/test-resources'
```
banterCZ, line 21: 🤔 I am not sure whether it is needed to archive the keystore at GHA. It is great that you can use the reproducible input in the local environment.
If you prefer to keep it at GHA, do we have to update our general `maven-test.yaml`? Isn't it possible to modify it only here and extract after there is more than one usage?

banterCZ (approved):
Thanks for generating test resources. It would break otherwise in the future.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
michal-rozehnal-w: I was also thinking about it. You are right — by default, tests are not executed in parallel, and I don’t see any override for this in GitHub Actions. I think it’s fine for now. If we decide to run them in parallel in the future, it should be reviewed across the entire project, because there are usually other obstacles.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
michal-rozehnal-w: I agree, it is safer. Done

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
michal-rozehnal-w: I also think this change should be ok. Moved into static block

.github/workflows/maven-test.yml
michal-rozehnal-w: Unfortunately, it seems this is not possible. At least I didn't find any solution and I am open to any ideas. Here are the things I considered:
- Add an extra step to the common job: This is not possible. When `uses:` is defined in a job, `steps:` cannot be used. Essentially, the logic of the common job is immutable.
- Use a separate job: In GitHub Actions, each job is isolated, so they don’t share any resources. Out of curiosity I tried it [here](https://github.com/wultra/signer-cloud/actions/runs/18713180399/job/53366350057?pr=135)

According to the documentation, GitHub Actions recommends using either `outputs:` for small text values or `actions/upload-artifact` and `actions/download-artifact` for files. In the case of self-hosted runners, it is also possible to define the same runner for multiple jobs, but this does not apply to our setup.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
```diff
@@ -0,0 +1,320 @@
...
+    public static IntTestResources prepare() throws Exception {
+        if (testResources == null) {
+            createResources();
+        }
+
+        return testResources;
+    }
+
+    private static void createResources() throws Exception {
+        if (Security.getProvider("BC") == null) {
```
michal-rozehnal-w, line 81: This is interesting, in GHA it works fine but locally for very first integration tests I was getting:
```
java.security.NoSuchProviderException: no such provider: BC

	at java.base/sun.security.jca.GetInstance.getService(GetInstance.java:83)
	at java.base/sun.security.jca.GetInstance.getInstance(GetInstance.java:206)
	at java.base/java.security.KeyPairGenerator.getInstance(KeyPairGenerator.java:301)
	at com.wultra.signercloud.server.IntTestUtils.generateKeystore(IntTestUtils.java:176)
	at com.wultra.signercloud.server.IntTestUtils.createResources(IntTestUtils.java:80)
	at com.wultra.signercloud.server.IntTestUtils.prepare(IntTestUtils.java:73)
```
It seems this setup is too fast locally. I added this to be sure that BouncyCastle is set.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
```diff
@@ -0,0 +1,320 @@
...
+        final var documentHashSha384 = toBeSignedSha384.getBytes();
+        final var documentSignatureSha384 = sign(userPrivateKey, documentHashSha384, "SHA384withECDSA");
+
+        final var signedDocumentSha384 = padesService.signDocument(dssDocument, signatureParamsSha384, new SignatureValue(SignatureAlgorithm.ECDSA_SHA384, documentSignatureSha384));
+        final byte[] signedDocumentBytesSha384;
+        try (final var stream = signedDocumentSha384.openStream()) {
+            signedDocumentBytesSha384 = stream.readAllBytes();
+        }
+
+        return new DocumentResources(
```
banterCZ: You created a builder but are not using it.

.github/workflows/maven-test.yml
banterCZ: It is up to you. IMHO, it is not necessary to store that file, but feel free to add it.

michal-rozehnal-w: I was also thinking about not uploading it and just having a flag parameter for persisting artifacts if needed. However, since these files are not large and GitHub automatically cleans them up after some time, I don’t think it should be an issue. If anything comes up, we can always remove this step anytime.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
michal-rozehnal-w: Hm, I’m wondering why this change wasn’t committed the first time, because I’m pretty sure I already did it. This time, I double-checked that the builders are used, and it’s all good now.

signer-cloud-server/src/test/java/com/wultra/signercloud/server/IntTestUtils.java
banterCZ: ~Interesting, I am getting the same issue with my test in a different repo.~ My bad, my issue is different.

PR #146, opened 2025-11-04
Fix #141: Add ojdbc driver

### LINKED ISSUE #141: Add ojdbc driver to the docker
### Task specification

Make it possible to run the Signer Cloud with the Oracle database.

korbelm:


Dockerfile
```diff
@@ -9,7 +9,10 @@ ENV JAVA_HOME=/opt/java/openjdk \
     LB_ARCHIVE_SHA256=184ffd609518091da42d6cd75e883b4f6ff1763cce8883e95fc99f7f05ca262d \
     PKG_RELEASE=1~jammy \
     LOGBACK_CONF=/opt/logback/conf \
-    TZ=UTC
+    TZ=UTC \
+    APP_PATH="/app/signer-cloud-server.war" \
+    EXTLIB_PATH="/app/extlib" \
+    OJDBC_VERSION="23.8.0.25.04"
```
banterCZ, line 15: 🤔 Why not, but don't we have it differently in the PAC? Will this be watched by dependabot?

banterCZ (approved):
Looks good.

Dockerfile
korbelm: no idea about dependabot, PAC has totally different structure....  this is based on nextstep

PR #171, opened 2025-12-08
Fix #170: Use lower-case object names for Oracle

### LINKED ISSUE #170: Fix SQL queries case sensitivity for Oracle Database
### Affected product(s) and version(s)

0.9.1

### Environment

_No response_

### Steps to reproduce, current result, expected result

For Oracle database any SQL query produces bad grammar exception. See log output

### Log output

action: createUpdateSigner, state: failed, errorMessage: PreparedStatementCallback; bad SQL grammar [SELECT "sc_signer"."ID" AS "ID", "sc_signer"."CSR" AS "CSR", "sc_signer"."STATUS" AS "STATUS", "sc_signer"."USER_ID" AS "USER_ID", "sc_signer"."CERTIFICATE" AS "CERTIFICATE", "sc_signer"."TIMESTAMP_CREATED" AS "TIMESTAMP_CREATED", "sc_signer"."CERTIFICATE_CHAIN" AS "CERTIFICATE_CHAIN", "sc_signer"."EXTERNAL_SIGNER_ID" AS "EXTERNAL_SIGNER_ID", "sc_signer"."TIMESTAMP_LAST_UPDATED" AS "TIMESTAMP_LAST_UPDATED", "sc_signer"."TIMESTAMP_CERTIFICATE_EXPIRATION" AS "TIMESTAMP_CERTIFICATE_EXPIRATION" FROM "sc_signer" WHERE "sc_signer"."EXTERNAL_SIGNER_ID" = ?]

michal-rozehnal-w:


docs/db/changelog/changesets/signer-cloud-server/1.0.x/20251205-rename-objects-to-lowercase.xml
```diff
@@ -0,0 +1,189 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.30.xsd"
+                   objectQuotingStrategy="QUOTE_ALL_OBJECTS">
```
michal-rozehnal-w, line 5: The `objectQuotingStrategy="QUOTE_ALL_OBJECTS"` setting is very important and should always be used in changesets. It ensures that case sensitivity is respected. For example, when we use `<createTable tableName="some_table">`, the table `some_table` (in snake case) is created the same way in all databases, regardless of the vendor. Without `objectQuotingStrategy="QUOTE_ALL_OBJECTS"`, the database dialect is respected: in Postgres it creates `some_table`, but in Oracle it creates `SOME_TABLE`. This case inconsistency can cause Oracle SQL queries to throw an ORA exception, indicating that the table does not exist.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20251205-rename-objects-to-lowercase.xml
```diff
@@ -0,0 +1,189 @@
...
+
+    <changeSet id="1" author="Michal Rozehnal" dbms="oracle">
```
michal-rozehnal-w, line 7: In short, this changelog is applied only to the Oracle database and renames all tables, columns, and sequences from uppercase (Oracle’s default) to lowercase.
It affects only Signer Cloud objects (those with the `sc_*` prefix). Liquibase tables DATABASECHANGELOG and DATABASECHANGELOGLOCK are not affected, as they are independent of the Spring Boot application.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/SchedulerConfig.java
```diff
@@ -43,6 +43,8 @@ public LockProvider lockProviderDefaultDataSource(final DataSource dataSource) {
                 JdbcTemplateLockProvider.Configuration.builder()
                         .usingDbTime()
                         .withJdbcTemplate(new JdbcTemplate(dataSource))
+                        .withTableName("\"shedlock\"")
```
michal-rozehnal-w, line 46: By default, Shedlock works correctly with any database dialect since it [executes](https://github.com/lukas-krecan/ShedLock/blob/master/providers/jdbc/shedlock-provider-jdbc-internal/src/main/java/net/javacrumbs/shedlock/provider/jdbc/internal/AbstractJdbcStorageAccessor.java) custom SQL queries. However, to keep things consistent (because we still have a Liquibase [changeset](https://github.com/wultra/signer-cloud/blob/develop/docs/db/changelog/changesets/signer-cloud-server/1.0.x/20250814-shedlock.xml) that creates the table), I converted the table name to lowercase and used quoting.

signer-cloud-server/src/main/java/com/wultra/signercloud/server/configuration/JdbcConfig.java
```diff
@@ -30,13 +36,36 @@
...
+    public JdbcCustomConversions jdbcCustomConversions() {
+        return new JdbcCustomConversions(List.of(
                 new DocumentVisualSignatureConverter.VisualSignatureToJsonConverter(),
                 new DocumentVisualSignatureConverter.JsonToVisualSignatureConverter()
-        );
+        ));
     }
+
+    @Bean
+    public Dialect dialect(final JdbcOperations jdbcOperations) {
```
michal-rozehnal-w, line 50: We need a custom `Dialect` because Spring Data JDBC combines two different approaches when generating SQL.

When annotations such as `@Table`, `@Column`, or `@Sequence` are used, Spring Data JDBC preserves the exact identifier and its case. In other words, if we declare `@Column("note") private String note`, it always generates `SELECT "note" ...` for all databases (Postgres, Oracle, etc.), fully ignoring database-specific dialect rules. Based on my understanding of [this comment](https://github.com/spring-projects/spring-data-relational/issues/768#issuecomment-2264737882), this is by design.

On the other hand, when an annotation is not used (just `private String note`), Spring Data JDBC generates the identifier according to the database dialect. For Postgres this results in `SELECT "note" ... `(lowercase), and for Oracle it becomes `SELECT "NOTE" ...` (uppercase).

The [documentation](https://docs.spring.io/spring-data/relational/reference/jdbc/mapping.html#mapping.conventions) recommends defining a custom `NamingStrategy`, but this does not solve the problem. The name returned from `NamingStrategy` is still passed to the dialect, and the dialect modifies it. For example, in Oracle, `NamingStrategy` produces `note`, but the dialect later transforms it to `"NOTE"`.

This is where the inconsistency appears. Liquibase creates objects according to the database dialect — for Oracle this means uppercase column names, e.g., `NOTE`. But in the application, fields annotated with `@Column("note")` remain lowercase, which results in `ORA-00942: table or view does not exist`, because in database there is `NOTE` but Spring Data JDBC produces query with `"note"` exactly according to name in annotation. Postgres is unaffected because its dialect preserves snake_case, and in H2 we explicitly use the parameter `CASE_INSENSITIVE_IDENTIFIERS=TRUE`.

That's why I updated the Liquibase configuration so that all objects are always generated in lowercase, regardless of database vendor, and implemented a custom Oracle Dialect (other 2 supported database works fine as mentioned above) so that lowercase identifiers are also used for fields without annotations.

It appears that there is no way to force Spring Data JDBC to apply dialect rules to fields that have explicit annotations. At least, I couldn’t find anything in the documentation indicating that this is possible.

An alternative approach is to annotate every property with `@Column`, which bypasses the dialect completely. However, this results in a lot of duplication in entity classes. Using a custom `Dialect` is a much cleaner solution.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20251205-rename-objects-to-lowercase.xml
```diff
@@ -0,0 +1,208 @@
...
+        <renameColumn tableName="shedlock" oldColumnName="NAME" newColumnName="name" />
+        <renameColumn tableName="shedlock" oldColumnName="LOCK_UNTIL" newColumnName="lock_until" />
+        <renameColumn tableName="shedlock" oldColumnName="LOCKED_AT" newColumnName="locked_at" />
+        <renameColumn tableName="shedlock" oldColumnName="LOCKED_BY" newColumnName="locked_by" />
+    </changeSet>
+    
+    <changeSet id="13" author="Michal Rozehnal" dbms="oracle">
+        <preConditions>
+            <sqlCheck expectedResult="1">
+                SELECT COUNT(*)
```
michal-rozehnal-w, line 151: It seems that sequences are handled differently by Liquibase or Oracle. The `<tableExists>` and `<columnExists>` elements respect case sensitivity, but `<sequenceExists>` does not (both `<sequenceExists sequenceName="SC_CALLBACK_EVENT_SEQ" />` and `<sequenceExists sequenceName="sc_callback_event_seq" />` return true). That’s why the SQL query is used in the precondition.

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20251205-rename-objects-to-lowercase.xml
```diff
@@ -0,0 +1,208 @@
...
+        <renameColumn tableName="sc_callback_event" oldColumnName="TIMESTAMP_NEXT_CALL" newColumnName="timestamp_next_call" />
+        <renameColumn tableName="sc_callback_event" oldColumnName="TIMESTAMP_DELETE_AFTER" newColumnName="timestamp_delete_after" />
+        <renameColumn tableName="sc_callback_event" oldColumnName="TIMESTAMP_RERUN_AFTER" newColumnName="timestamp_rerun_after" />
+        <renameColumn tableName="sc_callback_event" oldColumnName="ATTEMPTS" newColumnName="attempts" />
+        <renameColumn tableName="sc_callback_event" oldColumnName="IDEMPOTENCY_KEY" newColumnName="idempotency_key" />
+    </changeSet>
+
+    <changeSet id="3" author="Michal Rozehnal" dbms="oracle">
+        <preConditions>
+            <tableExists tableName="SC_DOCUMENT" />
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

docs/db/changelog/changesets/signer-cloud-server/1.0.x/20251205-rename-objects-to-lowercase.xml
```diff
@@ -0,0 +1,208 @@
...
+        <preConditions>
+            <tableExists tableName="shedlock" />
+        </preConditions>
+
+        <renameColumn tableName="shedlock" oldColumnName="NAME" newColumnName="name" />
+        <renameColumn tableName="shedlock" oldColumnName="LOCK_UNTIL" newColumnName="lock_until" />
+        <renameColumn tableName="shedlock" oldColumnName="LOCKED_AT" newColumnName="locked_at" />
+        <renameColumn tableName="shedlock" oldColumnName="LOCKED_BY" newColumnName="locked_by" />
+    </changeSet>
+
```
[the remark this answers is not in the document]
michal-rozehnal-w: Fixed

banterCZ (approved):
Thanks for the fix, but it is so sad that we haven't found a more elegant solution.

PR #222, opened 2026-05-19
Fix #221: Register Lombok annotation processor

### LINKED ISSUE #221: Fix: Support Lombok with Java 25
### Task specification

Lombok 1.18.x relies on implicit annotation processor discovery which is broken on Java 25 — the `logger` field (and other generated members) are not visible to `javac`:

```
cannot find symbol
  symbol: variable logger
```

Affects local development on Java 25 (any vendor). CI uses Java 21 and is unaffected.

**Root cause:** Starting with JDK 23, the default annotation processing policy changed to `none`.

**Fix:** Explicitly register Lombok via `annotationProcessorPaths` in the root `pom.xml`'s `maven-compiler-plugin` configuration.

See: wultra/tasklist#898 for parent epic.

vita-kotacka:
## Description

Registers Lombok as an explicit Maven annotation processor in the Maven compiler plugin configuration.

## Motivation and Context

JDK 23+ disables implicit annotation processor discovery by default, so Lombok must be configured explicitly for Java 25 compatibility.

Closes #221

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,18 @@
                 <artifactId>cyclonedx-maven-plugin</artifactId>
                 <version>${cyclonedx-maven-plugin.version}</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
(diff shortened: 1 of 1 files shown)


banterCZ (approved):

michal-rozehnal-w (approved):
Looks good

